PR #9, opened 2018-09-01
Add fedora-initramfs build target

alonamid:
This PR add a fedora-initramfs build target to the firesim-software build script.
The fedora-initramfs target generates an initramfs from a Fedora image so it can be run on Spike and other alternatives without block-device considerations. 
It assumes that firesim-software is found within the main firesim directory structure, and that the fedora-uniform workload has been built within firesim/deploy/workloads.


sagark:
that was fast

PR #17, opened 2019-01-25
Custom workloads

NathanTP:
This is a major overhaul of the workload generation system. It introduces custom workloads with a real build system and launch/testing framework. The first release will not introduce significant new workloads (all existing workloads still live in firesim/deploy/workloads), the porting effort will happen incrementally.

See the corresponding PR in firesim.git for full documentation.


sagark:
@NathanTP can you fix the merge conflicts against dev here?

NathanTP:
I merged dev this should fix any merge conflicts. This branch changes too many things for a real merge though, so I used the 'ours' strategy (ignore the existing branch).

README.md
```diff
@@ -1,10 +1,14 @@
-firesim-software
+Fire Marshal (firesim-software)
```
davidbiancolin: I propose we reduce this readme to a few lines that aggressively redirects to FireSim's docs.

full_test.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+# Enable extended globbing
+shopt -s extglob
+
+SUITE_PASS=true
+
+# Reset the test log
+echo "" > test.log
```
davidbiancolin: Why not just a new file?

test/incremental/README.md
```diff
@@ -0,0 +1,23 @@
+Tests the incremental building capabilities of the system. The test procedure
+is fairly specific, so it's best to always use the test.py script to test this. 
+
+Incremental builds means that sw\_manager won't rebuild images from the base
```
davidbiancolin: sw_manager -> marshal

wlutil/br/br.py
```diff
@@ -0,0 +1,107 @@
...
+        env = os.environ.copy()
+        env.pop('PERL_MM_OPT', None)
+        wlutil.run(['make'], cwd=os.path.join(br_dir, "buildroot"), env=env)
+
+    # Return True if the base image is up to date, or False if it needs to be
+    # rebuilt.
+    def upToDate(self):
+        # XXX There's something wrong with buildroots makefile, it throws an
+        # error and never reports being up to date.
+        # XXX DONT COMMIT THIS CHANGE YOUR DEFNITELY GOING TO FORGET TO UNDO THIS
```
davidbiancolin: This needs to be dealt with.

NathanTP: That comment was from when I had it hard-coded to never even check. Now it checks if there is a buildroot output, if not, it will rebuild. If there is some output, it will use it. This means it won't rebuild buildroot if you change some core thing in the buildroot distribution (like buildroot-config) or the buildroot overlay. To change those things, just run make in wlutil/br. Without this, every buildroot-based workload gets fully reproduced every time (10x+ build time).

The comment has been fixed, and I've added a readme to wlutil/br to clarify the situation. In practice, this isn't really an issue because even init scripts and stuff are handled through firemarshal overlays instead of buildroot overlays. We could (and maybe should at some point) replace buildroot with a pre-built image like we do for Fedora.

test/incremental/README.md
NathanTP: fixed

README.md
NathanTP: Agreed. Fixed.

full_test.sh
NathanTP: fixed. It now uses mktemp to create a random file name for each run and reports that to the user.

NathanTP:
I cleaned up some unnecessary complexity in tests and examples. We're down to 8k new lines instead of 26k :).

I think I've addressed all the open issues.

wlutil/launch.py
```diff
@@ -0,0 +1,96 @@
+import logging
+from .wlutil import *
+
+# The amount of memory to use when launching
+launch_mem = "16384"
```
davidbiancolin: This smells a little. Is this exposed somewhere?

NathanTP: This just matches the default firesim config. It's not exposed right now. I think stuff like this needs to go into a "board" abstraction for FireMarshal. I also think that the 'launch' capability may merge with FireSim at some point which would also require some new abstractions (a virtual agfi or something).

PR #18, opened 2019-02-13
Linux changes for compatibility with credit-based flow control

zhemao:
Since the initial two bytes of the header will be meaningful for flow control, need to zero them out in the driver.


zhemao:
For riscv-linux, should merge ucbbar-all-zero-header to ucbbar-all.

zhemao:
No longer necessary

PR #22, opened 2019-05-04
Fedora mirrors

NathanTP:
The primary fedora-riscv repo went down, plus they still haven't fixed the cert. This change adds backup mirrors but leaves the original repo in case it ever comes back (maybe we should remove it completely since it's not super secure or stable apparently?). I would also recommend pushing this to master on both firesim-software and firesim, the broken repo is a showstopper for anyone using fedora (if not, I'll make an issue so people can find the solution).


sagark:
can you PR this to master instead?

PR #25, opened 2019-05-22
Bump BR for gcc 8.3 support; point submodule at upstream

davidbiancolin:
See firesim/firesim#290.


NathanTP:
I fixed the initramfs thing. full_test.sh sort of passes (modulo the issues in #26 which are unrelated to this PR). I didn't test on firesim itself, just disk and initramfs based builds on spike and qemu.

davidbiancolin:
Thanks for bringing this home Nathan.

NathanTP:
This should probably merge against dev instead of master. I haven't really thought it out, but I suppose it makes sense to keep master in sync with firesim master.

davidbiancolin:
You're right, I carelessly opened it up against the wrong base branch.

Issue #30, opened 2019-06-05
guestmount failing on ubuntu


NathanTP:
Marshal may fail with a somewhat cryptic error on ubuntu systems. This is a known issue and it's being documented here for posterity:

Reproduce:
./marshal build test/command.json

Symptom (e.g.):

> subprocess.CalledProcessError: Command 'guestmount -a /scratch/nathanp/firesim/sw/firesim-software/images/br-base.img -m /dev/sda /scratch/nathanp/firesim/sw/firesim-software/disk-mount' returned non-zero exit status 1

In the detailed logs you'll see:
> libguestfs: error: /usr/bin/supermin exited with error status 1.

This is an issue with guestmount and ubuntu (marshal uses guestmount to manipulate target filesystems without requiring sudo). The bugs are documented [here](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/759725) and [here](https://bugs.launchpad.net/ubuntu/+source/libguestfs/+bug/1813662). The solution is to add read privileges to the kernel (e.g. sudo chmod +r /boot/vmlinuz-*).

NathanTP:
Closed because we can't fix it from mashal, it must be fixed by your sysadmin.

wallento:
Just to understand as this is a common error: Why would chipyard need access to the host kernel?

jerryz123:
Chipyard attempts to build firemarshal linux images by default in the chipyard setup script. Firemarshal requires guestmount with sudo to create the images. 

You can disable the default firemarshal builds in chipyard's setup script with some flag. I believe it is --skip-marshal.

wallento:
Yes, but what I mean is from the user experience point of view this error and the explanation is confusing.
Also, what you say actually is the change I was about to propose in a PR, because right now it doesn't sudo:

```
diff --git a/wlutil/wlutil.py b/wlutil/wlutil.py
index b68631f..ad8a02e 100644
--- a/wlutil/wlutil.py
+++ b/wlutil/wlutil.py
@@ -626,14 +626,14 @@ def mountImg(imgPath, mntPath):
             assert "nfs" not in fstype, f"Guestmount does not support {fstype} filesystems, change mount-dir to a non-NFS filesystem"
 
             pidPath = './guestmount.pid'
-            run(['guestmount', '--pid-file', pidPath, '-o', f'uid={uid}', '-o', f'gid={gid}', '-a', imgPath, '-m', '/dev/sda', mntPath])
[...]

Issue #20, opened 2019-03-04
Need sudo for build?


donggyukim:
Is there any way to build without sudo?

NathanTP:
I have experimental support for this on the 'standalone' branch of this repo. I just haven't had time to PR it. It will be merged in soon. One note here: you must install either guestmount or fuse-ext2 for it to work.

PR #31, opened 2019-06-05
Standalone

NathanTP:
Previous versions of firemarshal required sudo to operate. This wasn't intrinsic, just convenient. It also had some artificial dependencies on being a submodule of firesim. This PR removes these requirements so that firesim-software can work on its own.

The firesim submodule was a minor change that doesn't affect behavior significantly (just changes some warnings).

Getting rid of sudo was more involved. I had to switch to guestmount (a FUSE tool). Implications/changes:
1. Guestmount didn't like the way rsync and cpio worked (it handles permissions in the target system in a weird way). The result is that we don't manipulate file owners when constructing images, and guestmount seems to correctly convert from root to the host user.
2. We'll need to update firesim's machine-launch script to install libguestfs-tools. Anyone updating will need to install that.
3. People using this on Ubuntu hosts will need to jump through some extra hoops because guestmount doesn't play nice there. I've documented this at #30 for posterity.


NathanTP:
BTW: This PR addresses #20

marshal
```diff
@@ -6,9 +6,6 @@ import logging
 import wlutil
 import contextlib
 
-if 'RISCV' not in os.environ:
```
davidbiancolin: You don't depend on this variable anymore?

NathanTP: We never depended on it explicitly. I was just using it as a proxy to detect if you'd sourced sourcme. In reality, we just depend on the riscv toolchain. I just added back a check for that instead.

PR #41, opened 2019-07-08
Bump fedora

NathanTP:
This switches from a nightly build to one of the blessed stable builds. There are a few advantages:
1. The fedora team has tested the image more thoroughly
1. The image is hosted on a less shady website than the nightlys.
1. This is a more recent image, it is (hopefully) more stable and has the latest features.

I did only the minimum to our default linux config to get it working. Fedora actually provides linux configs with their stable releases, at some point I may look at trying to match those configs more closely (but this one works).

Testing:
- Tested all the fedora-based unit tests in marshal.
- Tested fedora-base on firesim


zohan180:
Hi, thanks for this branch, which is helpful in my case, one of the application I want to test out somehow only working with fedora 30. 

However, when I try to run on firesim, it hangs after type in log in info(user and password). With this setup: using firesim 1.6, with example_1config topology and quadcore-nic-config, runing fedora-uniform with fedora-base workload. Then, I tried these two: example_1config, quadcore-nic with basic linux uniform, br-base, which works fine; I also ran the fedora-uniform with no-net-config, quadcore-no-nic-config, it works fine, too. 

Have I miss something with my setup, or that could something not really been tested out yet? Thanks in advance for any info.

NathanTP:
@zohan180 : that issue was due to an incompatibility with the old kernel (4.15). Now that we bumped to the upstream linux 5.3, it should no longer hang.

PR #45, opened 2019-08-22
Update requirements for a bare FPGA developer AMI

tmagik:
Tested on ami-02b792770bf83b668, version 1.6.0


NathanTP:
I tried to handle the switch from master to dev from the github UI and it still brings in a bunch of history from master that we don't want. Can you try to cherry-pick or rebase or something to get on dev so it merges clean?

PR #46, opened 2019-08-24
Linux 5.3 Bump

NathanTP:
This is a fairly large change to get us off the fork of riscv-linux and onto the upstream standalone branch. The most significant change here is that we now create an initramfs that loads the icenet and block device drivers before booting the main system.

See firesim/firesim#334 for some documentation updates that will help in understanding the changes here.

A more complete list of changes:
* The default linux kernel is now (almost) upstream 5.3
    * There are still two bugs that we have to modify in a fork (both due to device tree differences). Hopefully these will go away as we get patches submitted upstream and/or fix rocketchip. For the moment we are on a fork that closely tracks upstream.
* Workload kernel configuration files are now fragments instead of full configurations (existing configurations will almost work, you just need to remove any lines about INITRAMFS and MODULES). Users should really work to update their configs to only contain explicitly needed changes to the riscv defconfig.
* The icenet and block device drivers are now factored out into 'boards/firechip/drivers'. They each have their own repos now.
* All workloads now boot into a custom initramfs in order to load the device drivers (especially the block device) before booting the target rootfs.


wlutil/build.py
```diff
@@ -186,23 +186,78 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
     if ret != 0:
         raise RuntimeError("Error while building workload")
 
+def makeDrivers(boardDir, linuxSrc):
+    driverDirs = pathlib.Path(boardDir).glob("drivers/*")
+    makeCmd = "make -C " + str(linuxSrc) + " ARCH=riscv CROSS_COMPILE=riscv64-unknown-linux-gnu- "
```
NathanTP: I do my own manual make command here instead of using the makefiles included with the drivers themselves. On the one hand, it's a bit safer, but on the other, it means any custom things done down the line in those Makefiles wont propagate to Marshal. Not sure what the best option is.

wlutil/build.py
```diff
@@ -186,23 +186,78 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
...
+        loadScript.unlink()
+
+    drivers = makeDrivers(boardDir, linuxSrc)
+
+    for driverPath in drivers:
+        shutil.copy(driverPath, initramfs_root)
+
+    with open(loadScript, 'w') as f:
+        for driverPath in drivers:
+            f.write("insmod /" + str(driverPath.name) + "\n")
```
NathanTP: In theory, instead of doing this I could create a '/lib/modules/LINUX_KERNEL_VERSION/' directory and have the kernel auto-load the modules without this loadScript. Unfortunately, tracking down LINUX_KERNEL_VERSION is a pain and this loadScript works just fine. I guess the other way would be more idomatic for linux though. Not sure the right approach.

wlutil/build.py
```diff
@@ -186,23 +186,78 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
...
 def makeBin(config, initramfs=False):
     log = logging.getLogger()
 
     # We assume that if you're not building linux, then the image is pre-built (e.g. during host-init)
     if 'linux-config' in config:
         linuxCfg = os.path.join(config['linux-src'], '.config')
-        shutil.copy(config['linux-config'], linuxCfg)
+        defCfg = os.path.join(config['linux-src'], 'defconfig')
+        # Create a defconfig to use as reference
+        if not os.path.isfile(defCfg):
```
NathanTP: I'm a bit concerned about caching this defconfig. It's possible that changes to either marshal or some custom linux kernel in a workload could change the defconfig and those changes wouldn't propagate. You'd need to manually delete the 'defconfig' file to get the changes. This should be super rare (when does the defconfig change?) but it's also unintuitive. The alternative is re-generating it ever time, which would add more time to the build. Open to suggestion here.

wlutil/build.py
```diff
@@ -186,23 +186,78 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
...
+            run(['make', 'ARCH=riscv', 'defconfig'], cwd=config['linux-src'])
+            shutil.copy(linuxCfg, defCfg)
+            with open(defCfg, 'a') as f:
+                f.write("CONFIG_BLK_DEV_INITRD=y\n")
+                f.write('CONFIG_INITRAMFS_SOURCE="' + str(initramfs_cpio) + '"\n')
+
+        # Create a config from the user fragments
+        kconfigEnv = os.environ.copy()
+        kconfigEnv['ARCH'] = 'riscv'
+        run([os.path.join(config['linux-src'], 'scripts/kconfig/merge_config.sh'),
```
NathanTP: In the future, I'm going to change this so that users can provide an arbitrary list of kconfig fragments and the configs can stack on parent workloads. I'm not doing this yet to keep the functional changes to a minimum in this PR.

wlutil/init.py
```diff
@@ -0,0 +1,36 @@
+import logging
+from .wlutil import *
+from .br import br
+
+# Initialize wlutil for the first time
+def oneTimeInit():
+    log = logging.getLogger()
+
+    # Build initramfs fs structure (git can't save these because reasons)
+    initramfs_dirs = ["bin", 'dev', 'etc', 'proc', 'root', 'sbin', 'sys', 'usr/bin', 'usr/sbin', 'mnt/root']
```
NathanTP: Git can't save empty directories, so I have to create them dynamically

wlutil/init.py
```diff
@@ -0,0 +1,36 @@
...
+    for d in initramfs_dirs:
+        if not (initramfs_root / d).exists():
+            (initramfs_root / d).mkdir(parents=True)
+
+    # Make busybox (needed for the initramfs)
+    log.info("Building busybox (used in initramfs)")
```
NathanTP: I could use the busybox from buildroot, but then we're beholden to them. I feel safer having a separate submodule just for this that we control. When I started messing with the buildroot config my images got broken (never figured out why).

wlutil/init.py
```diff
@@ -0,0 +1,36 @@
...
+    shutil.copy(wlutil_dir / 'busybox-config', busybox_dir / '.config')
+    run(['make', jlevel], cwd=busybox_dir)
+    shutil.copy(busybox_dir / 'busybox', initramfs_root / 'bin/')
+
+    # Apply linux patches to the default kernel
+    patches = list(board_dir.glob("*.patch"))
```
NathanTP: This step will go away once we get the patches upstreamed. I think it's a fair trade for not dealing with a custom fork. One downside will be that anyone using a custom kernel will need to address this themselves since we only patch the default kernel here.

wlutil/init.py
```diff
@@ -0,0 +1,36 @@
...
+
+    linuxRepo = git.Repo(linux_dir)
+    if linuxRepo.is_dirty():
```
NathanTP: This isn't great. I can't figure out a reliable way of detecting if the patches have already been applied. This makes 'marshal init' not exactly idempotent and results in this (somewhat confusing) warning on subsequent runs. It also bails out if you've customized the default kernel at all.

wlutil/init.py
```diff
@@ -0,0 +1,36 @@
+import logging
+from .wlutil import *
+from .br import br
+
+# Initialize wlutil for the first time
+def oneTimeInit():
```
NathanTP: I thought about building the base distros here since it can take a long time and we may as well batch up the high-latency initialization stuff if were running it anyway. It's not strictly needed but might improve the user experience (while slowing down build-setup significantly).

wlutil/initramfsRoot/init
```diff
@@ -0,0 +1,13 @@
+#!/bin/busybox sh
+
+/bin/busybox sh setupSystem.sh
+if [ $? != 0 ]; then
+  echo "Failed to setup system. Dropping to emergency shell."
+  exec /bin/busybox sh
```
NathanTP: With luck, this may make things a bit more robust for us when we boot things and the block device or early boot breaks or something.

wlutil/launch.py
```diff
@@ -53,7 +53,6 @@ def getQemuCmd(config, initramfs=False):
     if 'img' in config and not initramfs:
         cmd = cmd + ['-device', 'virtio-blk-device,drive=hd0',
                      '-drive', 'file=' + config['img'] + ',format=raw,id=hd0']
-        cmd = cmd + ['-append', '"ro root=/dev/vda"']
```
NathanTP: One side-effect here is that we don't need the 'root=' option anymore. The initramfs init script just looks in both places. The previous behavior was a bit of a hack anyway: we had to set CMDLINE_FALLBACK in the kernel to use the device-tree cmdline if present or the builtin if not, since Rocket doesn't provide one (amusingly, linux 5.2 has a broken CMDLINE_FALLBACK which cmdline.patch fixes)

wlutil/wlutil.py
```diff
@@ -9,17 +9,48 @@
 import shutil
 import psutil
 import errno
+import pathlib
```
NathanTP: When this was first written, pathlib wasn't available (Python 3.4) but now it is so I'm starting a slow switch to using pathlib everywhere. In this file I've just changed the low-hanging fruit that was directly affecting new code. I'll come through in a separate refactoring pass to fully switch over.

wlutil/build.py
```diff
@@ -186,23 +186,78 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
...
+            defCfg, config['linux-config']], env=kconfigEnv, cwd=config['linux-src']) 
+
+        setupBoardInitramfs(board_dir, config['linux-src'])
 
         if initramfs:
```
NathanTP: This is kinda confusing in both the code and in the user experience. We used to only have an initramfs if the user wanted their rootfs in an initrd. Now, we always have an initramfs. On disk-based systems it has the driver loading stuff, on diskless systems it has your rootfs. We still use 'initramfs' in most places to mean the diskless thing, but it's not as clear as it once was.

wlutil/initramfsRoot/setupSystem.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Load platform drivers
+sh /loadDrivers.sh
+
+# Mount the /proc and /sys filesystems.
+mount -t proc none /proc
+mount -t sysfs none /sys
+mount -t devtmpfs none /dev
+
+if [ -e /dev/generic-blkdev ]; then
```
NathanTP: This means that the kconfig and bootargs don't have to change based on the platform anymore

wlutil/build.py
tmagik: Anyone who's used linux on a desktop platform is probably familiar with the concept of using an initramfs to mount the real root filesystem. (https://fedoraproject.org/wiki/Dracut) or (https://wiki.debian.org/initramfs). Also along those lines, the bigger confusing/usability problem is use of the combine BBL+kernel image.

If you want to make the user experience easy, make it so unpatched Fedora & Debian kernels 'just work' in firesim. That means either porting U-boot, or refactoring the Verilator infrastructure so it can load an SBI implementation (bbl), kernel, and initramfs directly into memory and then add the addresses of the kernel and initramfs to the 'chosen' node. (see https://github.com/tmagik/freedom-u-sdk/blob/master/conf/uEnv.txt for a u-boot script example).

wlutil/build.py
NathanTP: After discussion, this has been switched to just all the driver's own makefile

wlutil/build.py
NathanTP: This now rebuilds the defconfig every time

wlutil/build.py
NathanTP: I changed the terminology to be 'nodisk' now instead of 'initramfs' to make the intent more clear (the old -i --initramfs option is still there for backward compatibility). Now the term 'initramfs' anywhere in the Marshal source refers to a literal initramfs.

The confusion around BBL+kernel will need to be dealt with in the documentation. I think there's some mention of it in the docs now, but we're going to try and update stuff a bit more in anticipation of an upcoming tutorial at the Micro '19 conference.

Unpatched kernels unfortunately won't work at the moment. We are in the process of submitting patches to get the known bugs fixed, but until then, we're stuck with the patches. As much as I'd love to stop messing around with kernel sources, we're just not there yet. This PR is a big step toward allowing that though.

I'm all for getting a proper bootloader going. If someone provides a working port that plays nice with Rocket/FireChip/etc. then I'll be happy to try and integrate it. BBL clearly has limitations, but we've been able to work around them so far.

tmagik:
Another suggestion for buildroot, at least.. 
```
# Put a getty on the serial port
#console::respawn:/sbin/getty -L  console 0 vt100 # GENERIC_SERIAL
console::respawn:-/bin/sh
```

wlutil/init.py
NathanTP: This has been removed in favor of a Linux fork that differs from upstream only in these two patches. We will strive to get off that and minimize changes.

wlutil/init.py
NathanTP: This is now gone in favor of the linux fork

NathanTP:
I've incorporated all the feedback from code review. It's ready for final approval.

- All local unit tests pass
- Fedora and buildroot both boot and can ping each other in a 2-node config (no further testing on firesim was done).
- Fedora diskless no longer works. This is likely a kernel bug. I will make an issue to track that.

wlutil/init.py
```diff
@@ -0,0 +1,19 @@
...
+    disk_dirs = ["bin", 'dev', 'etc', 'proc', 'root', 'sbin', 'sys', 'usr/bin', 'usr/sbin', 'mnt/root']
+    for d in disk_dirs:
+        if not (initramfs_dir / 'disk' / d).exists():
+            (initramfs_dir / 'disk' / d).mkdir(parents=True)
+
+    # Make busybox (needed for the initramfs)
+    log.info("Building busybox (used in initramfs)")
+    shutil.copy(wlutil_dir / 'busybox-config', busybox_dir / '.config')
+    run(['make', jlevel], cwd=busybox_dir)
+    shutil.copy(busybox_dir / 'busybox', initramfs_dir / 'disk' / 'bin/')
```
a0u: It's not obvious when a user forgets to run `firesim init`.  `firesim build` will still build images without raising an error.  For "disk" configurations, the absence of `busybox` from the final initramfs is only noticeable from a relatively cryptic `Failed to execute /init (error -2)` boot message.

NathanTP: ya, I was being stupid with the init command. I just made it idempotent and call init every time. Much simpler. No more need to call 'marshal init'.

NathanTP:
merge conflicts resolved. I also removed the 'init' command in favor of an idempotent init() function I call every time.

HengZhuo2:
Hi, thanks for this update, I think its pretty big and useful. However, I have some questions. I was trying to modify the linux-config and try to add GDB debug or ftrace feature, none of them works. Just wondering if that is something not considered supported now or I did not  do in the right way. Details as below.

For gdb, refer to https://01.org/linuxgraphics/gfx-docs/drm/dev-tools/gdb-kernel-debugging.html:
I added these to the linux-config:
```
 CONFIG_DEBUG_INFO=y
 # CONFIG_DEBUG_INFO_REDUCED is not set
 # CONFIG_DEBUG_INFO_SPLIT is not set
 # CONFIG_DEBUG_INFO_DWARF4 is not set
 # CONFIG_DEBUG_INFO_BTF is not set
 CONFIG_GDB_SCRIPTS=y
```
kernel build fine, but cannot launch due to kernel panic cannot load fs.
Then, as for the ftrace
```
CONFIG_FUNCTION_TRACER=y
CONFIG_FUNCTION_GRAPH_TRACER=y
# CONFIG_PREEMPTIRQ_EVENTS is not set
# CONFIG_IRQSOFF_TRACER is not set
# CONFIG_SCHED_TRACER is not set
# CONFIG_HWLAT_TRACER is not set
CONFIG_FTRACE_SYSCALLS=y
# CONFIG_TRACER_SNAPSHOT is not set
CONFIG_BRANCH_PROFILE_NONE=y
# CONFIG_PROFILE_ANNOTATED_BRANCHES is not set
# CONFIG_PROFILE_ALL_BRANCHES is not set
CONFIG_STACK_TRACER=y
# CONFIG_BLK_DEV_IO_TRACE is not set
CONFIG_UPROBE_EVENTS=y
CONFIG_DYNAMIC_EVENTS=y
CONFIG_PROBE_EVENTS=y
CONFIG_DYNAMIC_FTRACE=y
CONFIG_DYNAMIC_FTRACE_WITH_REGS=y
```
kernel build failed at iceblk-driver build.

NathanTP:
Hi Heng,

Could you please repost this question to our forum at https://groups.google.com/forum/#!forum/firesim

This is an old, closed, PR.

Thanks,
Nathan

PR #48, opened 2019-09-19
Add rocc test/example

NathanTP:
Here is the promised rocc example. We will need to add documentation to a different repo (currently marshal docs live in firesim). Some specific things I'd like feedback on:

1. **rocc.h**: I got this file from Howie who got it from someone who presumably got it from Schuyler. After committing and documenting this, it will become a de-facto standard. If anyone wants to take a stab at making a better one, be my guest. More generally, we can discuss if a template like this is really the best approach, or if we should provide more tooling to automate the process.

1. **README/Docs**: I don't claim to actually know how any of this works. I wrote the readme to the best of my ability, but someone like Howie or Albert might do a better job (or at least double check my explanations).

1. **using dummy_rocc**: I wanted to keep the example short and to the point, so I used the default 'dummy_rocc' accelerator that comes with spike. Most people will need to create a golden model and run a custom spike. The Sha3 example from the chipyard tutorial shows all that. I think after the tutorial, we might include that by default and use it as an additional example.


test/rocc/util.h
```diff
@@ -0,0 +1,135 @@
+// See LICENSE for license details.
```
zhemao: Is any of the code in this header actually used? It's included in syscalls.c, but none of the code here appears to be used anywhere else.

test/rocc/README
```diff
@@ -0,0 +1,28 @@
...
+# Usage
+Build with:
+    make
+
+Run on Spike with:
+    spike --extension=dummy_rocc rocc-test.riscv
+
+Run in Marshal:
+    ./marshal build test/rocc.json
+    ./marshal launch -s test/rocc.json
```
alonamid: "Run in Marshal" here is "Run in FireSim"?
(Assuming we're moving marshal up on level in the hierarchy, this is not trivial anymore. Need to differentiate between running in FireSim, running in Verilator, or running in QEMU/Spike)

test/rocc/README
```diff
@@ -0,0 +1,28 @@
...
+
+or:
```
alonamid: What is the difference between this option and the previous option?

test/rocc/README
NathanTP: No, this is "Run using Marshal" (using spike). I don't include instructions on running firesim in this readme, I suppose I could.

test/rocc/README
NathanTP: I'll clarify in the README. The first one runs the test but doesn't validate the outputs. The 'test' command automatically builds and validates outputs.

test/rocc/README
alonamid: Then maybe make this "Run in Marshal using Spike"?
(I think it would be useful to have instructions for bare-metal on FireSim and Verilator. We have the Explicit Verilator instructions in the software-simulation section, but then it's not clear how it relates to Marshal)

test/rocc/README
alonamid: Validate against what?

test/rocc/util.h
NathanTP: static_assert is used somewhere. The binary fails to link without including util.h due to a missing "static_assert".

test/rocc/README
NathanTP: This readme assumes you've read the documentation of FireMarshal. I don't want to repeat everything here. This is just a minimal description. It should explain how building and running a RoCC workload is different than other workloads. These one-liners are here just so you can copy/paste if you forgot the exact syntax.

alonamid: So it might be better to remove the last line (test) in order to not confuse anyone (if we're expecting people to have read the full documentation before hand, then they will know the `test` feature exist, and they don't need it in the readme).

NathanTP: I took another stab at the README. Let me know if it's more clear now.

NathanTP:
I made a few updates, this is ready for re-review:
- I updated to the latest dev branch
- Removed the committed generated files (and updated the gitignore)
- Updated the README

test/rocc/rocc-test.c
```diff
@@ -0,0 +1,48 @@
+// This test program uses the example ROCC accelerator that is included with
+// spike by default (called dummy-rocc).
+// Use the provided Makefile to build
+// Run with spike --extension=dummy_rocc rocc-test.riscv
+
+#include <stdio.h>
+#include <stdint.h>
+#include "rocc.h"
+
+// Old custom format:
```
alonamid: I'm not sure I understood this comment

NathanTP: Whoops! That was a note-to-self from when I was converting this code to Schuylers macros. I've removed it.

PR #50, opened 2019-10-01
Update Buildroot and switch to github mirror

NathanTP:
This PR updates the buildroot submodule to the latest stable release (2019.08) and switches to the github mirror (should be more stable than the official mirror). This also switches configuration management to use kfrags in a similar way to how linux configs are managed. This will simplify bumping buildroot in the future. I also dynamically detect the build environment and set that up properly, this will help decouple toolchain changes from buildroot (although we'll still need to bump BR frequently to stay ahead of the toolchain releases).

Testing:
* Ran basic buildroot tests locally on qemu
* Booted br on no-net config in firesim


wlutil/br/br.py
```diff
@@ -37,6 +38,48 @@
...
 
+def buildConfig():
+    """Construct the final buildroot configuration for this environment. After
+    calling this, it is safe to call 'make' in the buildroot directory."""
+
+    # Major/minor version of the linux kernel headers included with our
+    # toolchain. This is not necessarily the same as the linux kernel used by
+    # Marshal, but is assumed to be <= to the version actually used.
+    linuxHeaderVer = sp.run(
+      "{ echo '#include <linux/version.h>' ; echo 'LINUX_VERSION_CODE' ; } | riscv64-unknown-linux-gnu-gcc -E -xc - | tail -n 1",
```
a0u: It's possible to just exec `riscv64-unknown-linux-gnu-gcc` directly instead of using the shell.

Untested:
```python
linuxHeaderTest = """#include <linux/version.h>
LINUX_VERSION_CODE
"""
linuxHeaderVer = sp.run(['riscv64-unknown-linux-gnu-gcc', '-E', '-xc', '-'],
  input=linuxHeaderTest, stdout=sp.PIPE, universal_newlines=True)
linuxHeaderVer = linuxHeaderVer.stdout.readlines()[-1].strip()
```
There's probably a more Pythonic way to get the last line of stdout.

wlutil/br/buildroot-config
```diff
@@ -1,2983 +1,734 @@
...
+# BR2_KAGE_UNSCD is not set
+# BR2_KAGE_UNZIP is not set
+# BR2_KAGE_UPMPDCLI is not set
+# BR2_KAGE_URG is not set
+# BR2_KAGE_USSP_PUSH is not set
+# BR2_KAGE_UTIL_LINUX_AGETTY is not set
+# BR2_KAGE_UTIL_LINUX_BFS is not set
+# BR2_KAGE_UTIL_LINUX_BINARIES is not set
+# BR2_KAGE_UTIL_LINUX_CAL is not set
+
```
a0u: Is it worth disabling some unneeded packages to reduce the size of the default rootfs?

```
BR2_PACKAGE_APACHE
BR2_PACKAGE_APR
BR2_PACKAGE_CPPZMQ
BR2_PACKAGE_EXPAT
BR2_PACKAGE_PHP
BR2_PACKAGE_ZEROMQ
BR2_PACKAGE_ZMQPP
```

Do any FireSim workloads depend on these?

NathanTP: I'm not aware of any, I think minimizing the image is definitely worth it if we can get away with it. None of my unit tests use those, but I can't be sure about anyone else.

sagark: ISCA memcached workload depends on several of those. let's deal with this sort of thing later

wlutil/br/br.py
NathanTP: done

wlutil/br/buildroot-config
NathanTP: I'll leave it then

wlutil/br/br.py
```diff
@@ -37,6 +38,55 @@
...
+    linuxHeaderVer = linuxHeaderVer.stdout.splitlines()[-1].strip()
+
+    # Major/minor version of the linux kernel headers included with our
+    # toolchain. This is not necessarily the same as the linux kernel used by
+    # Marshal, but is assumed to be <= to the version actually used.
+    linuxMaj = str(int(linuxHeaderVer) >> 16)
+    linuxMin = str((int(linuxHeaderVer) >> 8) & 0xFF)
+
+    # Toolchain major version
+    toolVerStr = sp.run("riscv64-unknown-linux-gnu-gcc --version", shell=True,
```
a0u: This also doesn't need to be a shell command.  The added benefit is that `subprocess.run` raises `FileNotFoundError` (ENOENT) if the executable isn't in `PATH`, rather than having to explicitly check the return code.

NathanTP: done

PR #53, opened 2019-10-04
Convert RoCC macros to the .insn directive

a0u:
Now that a more modern binutils has arrived, the RoCC macros can be implemented more cleanly with the [`.insn` directive](https://sourceware.org/binutils/docs/as/RISC_002dV_002dFormats.html).

Mainly, the ability to use proper register constraints means that operands no longer have to be passed in fixed registers (`a0`, `a1`, `a2`), which should improve register allocation slightly.  As an additional benefit, the assembler will error if `funct` exceeds the immediate range ("illegal operands").

To encode immediates in the `rs1`/`rs2`/`rd` fields, we creatively reuse/abuse the `K` constraint, normally used for 5-bit unsigned CSR immediates.

I verified that `rocc-test.riscv` continues to pass on spike.

### DIFF
--- a/test/rocc/rocc-test.c
+++ b/test/rocc/rocc-test.c
@@ -13,11 +13,11 @@ int main() {
   // See rocc.h for details but roughly:
   // I_R_I => rd (Immediate), rs1 (Register), rs2 (Immediate)
   // opcode=0, rd=0 (no return value), rs1=x (register), rs2=2 (immediate),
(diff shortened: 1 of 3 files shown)


a0u:
Yes, I'm been using this for the SHA3 benchmark.  `rocc.h` will be synced in sha3-workload soon.

PR #60, opened 2019-10-07
Dev null

NathanTP:
There used to be some scary looking output on bootup about /dev/null not existing. This PR adds the /dev/null special file to the initramfs to eliminate the issue. It also cleans up the script that generates the special file cpio.


wlutil/initramfs/makeDevNodes.sh
```diff
@@ -0,0 +1,25 @@
...
+# Create the devNodes.cpio archive that contains special files for an initramfs
+
+orig_pwd=${PWD}
+tmp_dir=$(mktemp -d -t ci-XXXXXXXXXX)
+
+pushd $tmp_dir
+
+echo "Making Nodes"
+mkdir dev
+mknod -m 622 dev/console c 5 1
```
a0u: Interestingly, `fakeroot` works around `/tmp` being mounted with `nodev`.

PR #64, opened 2019-10-08
Do not unmount /dev after running initramfs

a0u:
The `/dev/null` errors and `dropbear` exiting prematurely after init are all manifestations of the same issue.  The syslog output from `dropbear` indicates what is happening:
```
Jan  1 00:00:12 buildroot authpriv.info dropbear[246]: Early exit: Failed to daemonize: No such device
```

Since buildroot's default `/etc/inittab` does not explicitly mount `/dev` (unlike `/proc` and `/sys`), `/dev` should not be unmounted after `/init`.  This is also why `CONFIG_DEVTMPFS_MOUNT` was originally required for the disk-based configuration before an initramfs was added.


a0u:
Note that [devNodes.cpio](../blob/dev/wlutil/initramfs/devNodes.cpio)  is unrelated to this, as the problem is with the rootfs `/dev`, not initramfs.

This also fixes the `/proc/sys/kernel/hotplug` error with mdev.

a0u:
This may have also solved our `poweroff` issue.  I was able to omit `-f` in spike and have it halt normally like QEMU.

PR #66, opened 2019-10-08
Only build one binary for nodisk

### LINKED ISSUE #62: building unneeded binaries
workloads that don't have a guest-init don't need to build a binary (especially if they are a parent of something else). We can remove the binary dependency for rootfs's for workloads that don't have a guest-init.

NathanTP:
When you build a workload in diskless mode you generally shouldn't need the normal disk-based binary, but Marshal will always build it right now. This fixes it so that the disk-based binary is only built in nodisk mode if you have a guest-init. It's still needed for workloads with a guest-init because those need to boot the disk-based system in order to construct the rootfs (which is then integrated into an initramfs).

Resolves #62

### DIFF
--- a/full_test.sh
+++ b/full_test.sh
@@ -94,6 +94,14 @@ if [ ${PIPESTATUS[0]} != 0 ]; then
   exit 1
 fi
 
+echo "Running inheritance test" | tee -a $LOGNAME
+./test/inherit/test.py >> $LOGNAME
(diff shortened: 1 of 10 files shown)


NathanTP:
FYI: I messed up the branching when I created this feature branch and accidentally included #65 This PR actually only changes three lines in build.py. Merge #65 first to get a clean file diff. Sorry.

PR #68, opened 2019-10-09
Skip coldplug module load

a0u:
This replaces the BusyBox `mdev` init script to omit the recursive sysfs scan, which takes a while in simulation and is unnecessary, since no modules are expected to be loaded after the initramfs.

```sh
find /sys/ -name modalias -print0 | \
	xargs -0 sort -u | \
	tr '\n' '\0' | \
	xargs -0 modprobe -abq
```

### DIFF
--- a/wlutil/br/buildroot-overlay/etc/init.d/S10mdev
+++ b/wlutil/br/buildroot-overlay/etc/init.d/S10mdev
@@ -1,0 +1,45 @@
+#!/bin/sh
+#
+# Run the mdev daemon
+#
+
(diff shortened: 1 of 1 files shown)


NathanTP:
I feel that this warrants enough discussion that we might consider holding off until the release to include it. In the meantime, I think including it in the tutorial branch is fine.

a0u:
To be safer, the original code is now kept around but skipped if `/lib/modules` does not exist.  I'll leave this open for now, and we can deal with it after the tutorial.

PR #69, opened 2019-10-10
Add Linux timebase-frequency divisor hack

a0u:
This creates the illusion that the `mtime` RTC is slower than the actual hardware frequency, which makes simulations on F1 feel more responsive.

This currently affects only "br-base" and descendants.  We should consider reverting the Kconfig change post-tutorial and investigate a cleaner solution in the long term.

### DIFF
--- a/workloads/br-base/linux-config
+++ b/workloads/br-base/linux-config
@@ -14,6 +14,11 @@ CONFIG_RISCV_BASE_PMU=y
 CONFIG_RISCV_ISA_C=n
 
 #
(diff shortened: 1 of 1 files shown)


sagark:
I think this shouldn't end up on dev even for now. Can we split off to a branch?

a0u:
Yes, once the sha3-workload is bumped and the other PRs are merged, we can start branching.

a0u:
Pushed to the tutorial branch instead.

PR #81, opened 2019-10-22
Multiple workdirs

NathanTP:
Allows user to have multiple working directories, and for those directories to be anywhere in the system. The main change is putting the standard builtin bases in boards/firechip/base-workloads and including that including that in the search path for workloads by default. This means we don't need to symlink or copy things like br-base.json.

The main motivation here is to allow things like chipyard/software/tests/ to include standard bare-metal tests (like the isa-test and other standard chipyard tests). It also will allow projects to define all their workloads in a standalone repo without needing the 'install.sh' script that, e.g., sha3 uses. Finally, this paves the road to possibly including different bases for different boards (e.g. a tapeout might need standard init scripts or configure the NW differently than firechip would).

Internally, this also identifies workloads by their file name only, not by the full path to the filename. A warning is issued if multiple workloads have the same file name.

Allows marshal to be called from anywhere in the system (it no longer has any implicit dependencies on pwd).


NathanTP:
One question for reviewers: How do you feel about putting all the stuff currently in wlutil/br/buildroot-overlay into br-base/overlay? br-base represents the board's (firechip's) configuration, and most of that stuff is firechip-specific (mostly networking). Fedora has all that stuff in fedora-base.json. The benefit would be in reducing complexity in the wlutil/br module, and being more idiomatic for marshal. It would also make it easier to add different boards in the future.

.gitmodules
```diff
@@ -17,5 +17,5 @@
 	path = boards/firechip/drivers/iceblk-driver
 	url = https://github.com/firesim/iceblk-driver.git
 [submodule "workloads/sha3"]
```
a0u: For consistency, the submodule name should follow the path change.

NathanTP: wow, git really doesn't fix that? I even used 'git mv' to do it. Whatever. It's fixed now.

NathanTP:
I also made it so you can do things like './marshal build br-base.json' even if you're not in the boards dir. Basically, there are multiple search paths now (kinda like how PATH works)

NathanTP:
fixed

.gitmodules
a0u: It's hard to say whether this omission is deliberate or not because renaming might break existing `.git/config` settings under the old name, although I have not checked.  But the discrepancy would be confusing to someone running `git config` expecting to use the new path.

PR #89, opened 2019-10-29
Detect Changes in buildroot submodule

NathanTP:
Buildroot base image now gets rebuilt if the buildroot submodule commit hash changes, or if it has any uncommitted changes. A warning is issued if there are uncommitted changes to make it more clear what is happening.


NathanTP:
superseded by #93

PR #93, opened 2019-10-31
Marshal init

NathanTP:
This makes it so that Marshal only depends on submodules if it actually needs them. Useful warnings are issued if a dependency is missing. It also cleans up/overhauls a lot of dependency checking throughout the system. In particular, linux and buildroot are now considered uptodate unless:
- They have uncommitted changes (in tracked files)
- Their git commit has changed

All workloads now depend on the toolchain version, so changes to the toolchain will force a rebuild.


wlutil/br/br.py
```diff
@@ -82,22 +64,19 @@ def buildConfig():
     # Our config fragment, specifies differences from the default
     marshalKfrag = br_dir / 'buildroot-config'
 
-    # We're just borrowing linux's merge_config.sh helper since br uses the same make system
-    # This doesn't actually depend on any details of this linux kernel
-    mergeScript = str(pathlib.Path(wlutil.linux_dir) / 'scripts' / 'kconfig' / 'merge_config.sh')
+    mergeScript = br_dir / 'merge_config.sh'
```
NathanTP: I had to copy this out of linux because the the buildroot image might be built before the binary (where the linux submodule check happens). This is also probably a bit cleaner anyway.

wlutil/build.py
```diff
@@ -20,20 +20,21 @@ def load_tasks(self, cmd, opt_values, pos_args):
         config = {'verbosity': 2}
         return task_list, config
 
-# Checks if the linux kernel used by this config needs to be rebuilt
```
NathanTP: This is no longer needed because we now have a gitCheckStatus() function

wlutil/wlutil.py
```diff
@@ -71,6 +74,20 @@
...
+    def __init__(self, path):
+        self.message = "Dependency missing or not initialized " + \
+                str(path) + \
+                ". Do you need to initialize submodules?"
+        self.path = path
+
+    def __repr__(self):
+        return "Submodule Error: \n" + self.message
+
+    def __str__(self):
```
NathanTP: stack overflow makes it sound like __str__ should default to __repr__, but that wasn't working for me. I'm not sure what's going on with that.

wlutil/wlutil.py
```diff
@@ -84,13 +101,6 @@ def initialize():
         if not (initramfs_dir / 'disk' / d).exists():
             (initramfs_dir / 'disk' / d).mkdir(parents=True)
 
-    # Make busybox (needed for the initramfs)
```
NathanTP: This is now handled as a task in the doit graph

wlutil/wlutil.py
```diff
@@ -268,3 +278,144 @@ def copyImgFiles(img, files, direction):
...
+        for any reason (e.g. dirty==True or init==False). 0 otherwised.
+
+    This is primarily useful as an input to doit's config_changed() updtodate
+    helper which considers a workload not uptodate if some string or dictionary
+    has changed since the last time it ran. The 'sha' or 'rebuild' fields will
+    change if the repo has changed (or we can't tell if it's changed)."""
+
+    log = logging.getLogger()
+
+    if submodule == "":
```
NathanTP: this is useful for having optional submodules (e.g. linux). We can just pass a default value that is trivially uptodate to represent that we don't depend on something.

wlutil/wlutil.py
```diff
@@ -268,3 +278,144 @@ def copyImgFiles(img, files, direction):
...
+
+    submodule: Path to the submodule to check. This check will be skipped if
+        the empty string is passed (the empty string is always uptodate)
+
+    Fields:
+    'sha'  : latest git commit hash (or "" if not initialized)
+    'dirty': boolean indicating whether there are uncommited changes (uninitialized repos are considered dirty)
+    'init' : boolean indiicating if the repository has been initialized
+    'rebuild' : A random number if the repo should be considered not up to date
+        for any reason (e.g. dirty==True or init==False). 0 otherwised.
```
NathanTP: kinda hacky, but really useful. The config_changed() helper just checks to see if something is different, a random number is effectively guaranteed to be different.

wlutil/wlutil.py
```diff
@@ -311,3 +321,144 @@ def copyImgFiles(img, files, direction):
...
+
+    It is safe to call this on something that is not actually a submodule. In
+    that case, this will simply check if the directory is empty or not.
+
+    s: Pathlib path to submodule
+
+    raises SubmoduleError if submodule not ready 
+    """
+    
+    if not s.exists() or not any(os.scandir(s)):
```
a0u: This doesn't cover the case that a submodule is uninitialized but the directory is non-empty, but the subsequent `make` error is intuitive enough to figure out what went wrong.

Issue #35, opened 2019-06-12
Rootfs resizing


NathanTP:
If you specify a lot of data in an overlay, you can run out of filesystem space in the rootfs. Current images are also much bigger than needed to make this less likely.

Feature request: Dynamically resize the rootfs based on the workload, leave a more modest margin for manual data. This will need an option in the workload config to increase that margin (e.g. if tests produce lots of data).

NathanTP:
Additionally, there is a resize.sh in wlutil/fedora/ that should serve as a workaround. However, centos 7 is stuck on an old version of e2fsprogs that can't work with fedora images. It should (ironically) still work on buildroot-based images.

A workaround proposed by aleksandar.pajkanovic@gmail.com ([here](https://groups.google.com/forum/#!topic/firesim/FbY3Ju03wIc)) should work for now:

1. make sure that the image is not running, shut it down if it is
1. go to the folder where image is stored and execute `qemu-img resize stage4-disk.img +1G` (replace, of course, the name of the image file and the size you want to add)
1. start qemu vm again
1. login
1. execute: `resize2fs /dev/vda`
1. run `df -h` to see that the disk size has increased

NathanTP:
resolved by #94

PR #94, opened 2019-11-02
Rootfs sz

NathanTP:
This adds the 'rootfs-size' option to workloads that allows you to explicitly size your images. Since you can now set it to whatever you need, the default is to resize images to have 256MiB of free space. This is a massive reduction in disk usage for FireMarshal (all buildroot images go from 2GiB to ~300MiB).

Fedora builds require a more recent version of e2fsprogs than comes with centos7. I had to manually install a more recent version (built from source). Before bumping Chipyard/FireSim, we'll need to modify the machine-launch-script.

Addresses #35 

@abejgonzalez


abejgonzalez:
Do you have the changes to the `machine-launch-script` somewhere? I can go ahead and try this and verify it works if the extra install is quick/easy

NathanTP:
it should work without changes to machine-launch-script so long as you don't use fedora. The new version of e2fsprogs is only needed for Fedora. I don't have that change anywhere yet.

docs/source/config.rst
```diff
@@ -226,6 +226,16 @@ should be absolute with respect to the workload rootfs. Files will be placed
 together in the output directory. You cannot specify the directory structure of
 the output.
 
+rootfs-size
+^^^^^^^^^^^^^^^^^
+The desired rootfs size (in human-readable units, e.g. "4GB"). This number must
```
a0u: The docs imply decimal units ("GB"), but `humanfriendly.format_size()` is called with `binary=True`.  The test JSON uses IEC binary units ("GiB").  Maybe this should be clarified?

NathanTP: I can make the example and the docs consistent, but it actually doesn't matter. humanfriendly .parse_size reads things correctly (GiB will be binary and GB will be decimal). It always prints in binary units and the suffix will indicate that (format_size is only for logging purposes). Users are free to do whatever they please.

PR #95, opened 2019-11-04
Update SHA3 Workdir Path

abejgonzalez:
This adds SPEC17 and CoreMark workloads to FireMarshal. Additionally, it updates the SHA3 workloads to work properly with relative paths. 

Depends On:
- #94


abejgonzalez:
This still needs to be tested with the rootfs work, so this PR is put up early in case people want to review things earlier.

NathanTP:
The spec workload looks OK, although I'll have to run through it to check more. However, I'm not sure how I feel about adding this to FireMarshal directly. I think it's probably better to add this to chipyard. My current plan is to wait for #93 to go through, and then:

- move marshal to chipyard/tools/firemarshal
- move sha3 to chipyard/software or the sha3 generator repo (not sure yet)

I think it might be better to put spec in, e.g., chipyard/software/spec2017 and keep firemarshal/example-workloads for demonstrations of FireMarshal specifically.

abejgonzalez:
If FireMarshal is moved to Chipyard then the PR in FireSim to bump FireMarshal + run the setup steps will be void (needs to move into someplace to update firemarshal stuff and install it's dependencies). Im fine with this stalling for a bit though.

NathanTP:
may as well merge the firesim one, no telling how long until I actually get around to moving stuff. I'm just saying you may as well PR this spec submodule into chipyard/software instead of marshal.

abejgonzalez:
Ok. I switched this to just update the SHA3 repository with the relative path changes. https://github.com/ucb-bar/chipyard/pull/326 is the Chipyard commit that does what you recommended.

PR #96, opened 2019-11-06
Sim params

NathanTP:
This adds a 'cpus' and 'mem' option to workloads so you can specify functional simulation parameters. It doesn't affect the 'install' command (yet).


wlutil/launch.py
```diff
@@ -84,6 +89,7 @@ def launchWorkload(cfgName, cfgs, job='all', spike=False, interactive=True):
         # Run the base image
         config = cfgs[cfgName]
  
+    print(config)
```
a0u: This seems like a vestigial debugging statement.

NathanTP:
Missed another leftover print. Hopefully got em all now.

PR #97, opened 2019-11-11
Run a machine

NathanTP:
Marshal guestmount mode stopped working on some machines after the sudo changes in #71. This fixes it.

### DIFF
--- a/wlutil/wlutil.py
+++ b/wlutil/wlutil.py
@@ -227,17 +227,17 @@ def waitpid(pid):
 
 if sp.run(['/usr/bin/sudo', '-ln', 'true'], stdout=sp.DEVNULL).returncode == 0:
     # User has passwordless sudo available, use the mount command (much faster)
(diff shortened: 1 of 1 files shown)


wlutil/wlutil.py
```diff
@@ -108,6 +108,9 @@ def initialize():
     Is safe and fast to call every time you load the library."""
     log = logging.getLogger()
 
+    if not os.path.exists(mnt):
+        run(['mkdir', mnt])
```
a0u: Are the intermediate path components of `mnt` guaranteed to already exist?  The `-p` flag is necessary otherwise.
You can also use [`os.mkdir()`](https://docs.python.org/3/library/os.html#os.mkdir) or [`os.makedirs()`](https://docs.python.org/3/library/os.html#os.makedirs)  here instead of an external command.

NathanTP: fixed.

PR #100, opened 2019-11-13
Use RVC instructions in buildroot

jerryz123:
The `defconfig` has this set, but the `linux-config` for buildroot unsets it for some reason.


NathanTP:
if it's in defconfig, you should be able to remove it completely from the br-base linux-config. I think that would be clearer (stuff in the workload kconfigs should be things we explicitly need that differ from defconfig).

Also, can you do the same to fedora-base?

jerryz123:
> didn't I already approve this?

I force pushed to fix the commit author metadata

PR #101, opened 2019-11-14
Switch to github mirror for busybox

NathanTP:
The official busybox github server is janky and breaks all the time. I just mirrored it on the ucb-bar github org.

This PR also cleans up some stuff related to building busybox.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/buildroot/buildroot.git
 [submodule "wlutil/busybox"]
 	path = wlutil/busybox
-	url = https://git.busybox.net/busybox/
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -9,7 +9,7 @@
 	url = https://github.com/buildroot/buildroot.git
 [submodule "wlutil/busybox"]
 	path = wlutil/busybox
-	url = https://git.busybox.net/busybox/
+  url = https://github.com/ucb-bar/busybox.git
```
sagark: there's something funky with the indentation here

NathanTP: fixed

PR #107, opened 2019-11-15
Marshal config

NathanTP:
This is an extensive refactor of the global configuration options in FireMarshal, as well as a more complete switch to pathlib from os.path. This pays down significant technical debt and prepares us for more global configuration changes (e.g. adding a 'fsim-dir' option to allow easy location of install targets).

The pathlib thing is just more convenient/concise and was used somewhat sporaticaly in the code. Now if you ever encounter an object that represents a path, it will be a pathlib object.

This was tested using full_test.sh.


NathanTP:
For reviewing: Probably best to read the docs changes and wlutil/wlutil.py first, then look at everything else.

wlutil/br/br.py
```diff
@@ -121,22 +121,22 @@ def generateBootScriptOverlay(script, args):
...
-            wlutil.run(['rm', scriptDst])
+            scriptDst.unlink()
             # Create a blank init script because overlays won't let us delete stuff
             # Alternatively: we could consider replacing the default.target
             # symlink to disable the firesim target entirely
-            wlutil.run(['touch', scriptDst])
+            scriptDst.touch()
         
-        wlutil.run(['chmod', '+x', scriptDst])
+        scriptDst.chmod(0o777)
```
a0u: Is there a reason why the script has to be world-writable instead of `0755`?

wlutil/fedora/fedora.py
```diff
@@ -52,27 +52,26 @@ def generateBootScriptOverlay(self, script, args):
...
-            wlutil.run(['rm', scriptDst])
+            scriptDst.unlink()
             # Create a blank init script because overlays won't let us delete stuff
             # Alternatively: we could consider replacing the default.target
             # symlink to disable the firesim target entirely
-            wlutil.run(['touch', scriptDst])
+            scriptDst.touch()
         
-        wlutil.run(['chmod', '+x', scriptDst])
+        scriptDst.chmod(0o777)
```
a0u: Same question as above.

docs/Makefile
```diff
@@ -16,4 +16,11 @@ help:
...
 %: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
\ No newline at end of file
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+# This will build the docs and launch a server. To access remotely you can use
+# SSH port forwarding:
+#     ssh -L 8000:localhost:8000 HOST
+# where HOST is the server actually hosting the docs
+server: html
```
a0u: Minor nit: `server` should be declared a phony target.

wlutil/br/br.py
NathanTP: because unix permissions confuse me :). I switched to 755.

docs/Makefile
NathanTP: fixed

NathanTP:
Fixed review comments. Ready for re-approval

PR #112, opened 2019-12-10
[install] use context paths and convert outputs to posixpaths

### LINKED ISSUE #110: [dev]  install.py needs to use ctx['key']
For the `root_dir` and `wlutil_dir` values, these need to be `ctx['root-dir']` and `ctx['wlutil-dir']` respectively. Otherwise, install will error since `root_dir` does not exist.

abejgonzalez:
Fixes #110 and also fixes bugs with outputs specified in during the `./marshal install ...` command.

### DIFF
--- a/wlutil/install.py
+++ b/wlutil/install.py
@@ -4,12 +4,6 @@ import json
 from pathlib import Path
 from .wlutil import *
 
-# firesim workloads directory
-try:
-    fsWork = (Path(root_dir) / "../../deploy/workloads").resolve()
-except:
-    fsWork = None
-
 readmeTxt="""This workload was generated using firesim-software. See the following config
 and workload directory for details:
 """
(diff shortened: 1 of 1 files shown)


wlutil/install.py
```diff
@@ -21,6 +15,7 @@ def fullRel(base, target):
     return os.path.relpath(str(target), start=str(base))
 
 def installWorkload(cfgName, cfgs):
+    fsWork = ((Path(getCtx()['root-dir']) / "../../deploy/workloads").resolve())
```
NathanTP: this should use getOpt instead of getCtx()[] (that is the consistent syntax that I've used elsewhere in the code).

wlutil/install.py
```diff
@@ -36,7 +31,7 @@ def installWorkload(cfgName, cfgs):
 
     # Path to dummy rootfs to use if no image specified (firesim requires a
     # rootfs, even if it's not used)
-    dummyPath = fullRel(fsTargetDir, Path(wlutil_dir) / 'dummy.rootfs')
+    dummyPath = fullRel(fsTargetDir, Path(getCtx()['wlutil-dir']) / 'dummy.rootfs')
```
NathanTP: use getOpt()

PR #116, opened 2019-12-18
Support for workload-specific riscv-pk

NathanTP:
Also strengthens some of the version checking for riscv-pk.

FYI:
@dayeol

### DIFF
--- a/docs/source/workloadConfig.rst
+++ b/docs/source/workloadConfig.rst
@@ -154,6 +154,10 @@ only the minimum necessary changes for your workload. In particular, you should
 avoid specifying a custom initramfs since Marshal provides it's own for loading
 platform drivers.
 
+pk-src
+^^^^^^^^^^^^^^^
+Path to riscv-pk source directory to use for this workload. This provides the bootload. Defaults to the riscv-pk submodule included at ``riscv-pk/``.
+
 host-init
 ^^^^^^^^^^^^^^
 A script to run natively on your host (i.e., them machine where you
--- a/docs/source/workloadConfig.rst
+++ b/docs/source/workloadConfig.rst
@@ -156,7 +156,7 @@ platform drivers.
 
 pk-src
 ^^^^^^^^^^^^^^^
-Path to riscv-pk source directory to use for this workload. This provides the bootload. Defaults to the riscv-pk submodule included at ``riscv-pk/``.
+Path to riscv-pk source directory to use for this workload. This provides the bootloader (bbl). Defaults to the riscv-pk submodule included at ``riscv-pk/``.
 
 host-init
 ^^^^^^^^^^^^^^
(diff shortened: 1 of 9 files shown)


docs/source/workloadConfig.rst
```diff
@@ -154,6 +154,10 @@ only the minimum necessary changes for your workload. In particular, you should
 avoid specifying a custom initramfs since Marshal provides it's own for loading
 platform drivers.
 
+pk-src
+^^^^^^^^^^^^^^^
+Path to riscv-pk source directory to use for this workload. This provides the bootload. Defaults to the riscv-pk submodule included at ``riscv-pk/``.
```
a0u: s/bootload/bootloader/

test/pk-src/copy-src.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+if [ ! -d riscv-pk ]; then
+  rsync --exclude ".git" -r ../../riscv-pk .
```
a0u: For the sake of completeness,
```sh
mkdir -p riscv-pk
git -C ../../riscv-pk archive HEAD | tar -x -f - -C riscv-pk
```
is potentially faster if the working tree is dirtied by build objects, which `git archive` would avoid copying.
This matters less here but could be significant for the linux-src test.

test/pk-src/copy-src.sh
```diff
@@ -0,0 +1,5 @@
...
+  patch riscv-pk/bbl/bbl.c < test.patch
```
a0u: I wish there was a cleaner testing method than maintaining out-of-tree patches that are potentially brittle.
To lessen the problem, `copy-src.sh` should begin with `set -e`, so that marshal would terminate at an earlier point if the patch failed to apply.  (Same goes for the linux-src test.)

a0u: Actually, I guess that would be redundant since the exit code would be inherited from the patch command being the last one in the script, but `set -e` is a good precaution anyway.

test/pk-src/copy-src.sh
NathanTP: I'll change it if it becomes a problem. My way is easy to guess its function, and requires a single man-page lookup if it's still not clear. I don't even know what exactly the suggested command does and I doubt future readers would either. It also doesn't work when copy/pasted (I don't understand it enough to fix) and the ' - ' option is not easily discovered in the tar man page (I still don't know what it does, is it BASH magic?). Anyway, "premature optimization is the root of all evil" and all that.

test/pk-src/copy-src.sh
NathanTP: 'set -e' is a good idea. I added that. 

I'm open to suggestions for a more reliable way to test a non-standard bbl, but this way should be pretty stable (it's a one-line change that doesn't depend on much). Anyway, it will only stop working if we bump riscv-pk, and we would run this test if we did that anyway.

test/pk-src/copy-src.sh
a0u: `git archive` creates a tarball from the specified revision, similar to GitHub releases.
`-` is the argument to `-f` and, by Unix convention, is shorthand for stdin.

PR #117, opened 2019-12-18
Allow for custom Qemu

NathanTP:
We already support custom spike, this just adds similar support for Qemu. This was motivated by the keystone project that needs a custom Qemu.

@dayeol


test/qemu/replace.py
```diff
@@ -0,0 +1,18 @@
...
+# for this feature is to use a fork or patched version of the qemu source and
+# rebuild it in your workload (that is just too slow for a unit test like
+# this).
+
+import shutil
+sysQemu = shutil.which("qemu-system-riscv64")
+shutil.copy(sysQemu, "testQemu")
+
+with open("testQemu", "rb") as input_file:
+    content = input_file.read().replace(b"QEMU emulator version", b"HELLO FROM QEMU TEST\n")
```
a0u: If the intent is to check that a different executable is run, it is not necessary to patch the QEMU binary.  Just make `testQemu` a wrapper script:
```sh
#!/bin/sh
echo "HELLO FROM QEMU TEST"
exec qemu-system-riscv64 "$@"
```

NathanTP: OK mister "do it the obvious and easy way", I'll do it your way. My way makes me look clever, which is totally better than being all stable and easy to understand.

a0u:
Just mark 5ca1833745a531592efc27735ea419e7747948e0 ("Simplify qemu test.") as a `fixup` in `git rebase -i` to squash it with b93c0827c4ccd016e44e5697d1d34ebf4a368b0b ("Add unit test for custom qemu binary").

NathanTP:
Well that was deeper down the git rabbit hole than I was expecting. Hopefully I did that right. It should be good now?

PR #118, opened 2019-12-20
Doit opts

### LINKED ISSUE #104: doit.db location
Doit uses an internal database for tracking changes and dependencies. By default this goes to PWD. This is potentially a problem when you start calling marshal from different places because it leaves all these .doit.db files everywhere. We should probably figure out how to put them in specific places.

NathanTP:
Add doit configuration options to marshal's config system. Most importantly, this changes the location of the doit.db file (it now defaults to wlutil/generated/marshaldb.*. Users can also override this option if they need (unlikely, but possible).

This resolves #104


a0u:
The `doItOpts` branch includes history from #116.  Could you merge the other PR first so this one only has commit aafce559f1775d852f36363e646eb4604fbf5888?

NathanTP:
Whoops. Git strikes again. Is it good now that #116 is merged or do I need to fix something else?

PR #120, opened 2020-01-22
Update CentOS dependencies

a0u:
`bc` is required for Linux kernel builds (specifically for header file generation).

### DIFF
--- a/centos-requirements.txt
+++ b/centos-requirements.txt
@@ -6,7 +6,7 @@ libguestfs-tools
 dtc
 makeinfo
 texinfo
-yac
+yacc
 bison
 flex
 expat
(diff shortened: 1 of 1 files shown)


centos-requirements.txt
```diff
@@ -14,3 +14,4 @@ expat-devel
 glib-devel
 gtk2-devel
```
a0u: On a similar note, could `glib-devel` and `gtk2-devel` be removed?  These are QEMU dependencies, which should already be handled elsewhere.

Actually, `gtk2-devel` is overkill for a headless system - `libfdt-devel pixman-devel zlib-devel` is the minimum for QEMU.

NathanTP: This list is hard to test so I don't really know what is needed for what. The intention is to only have requirements for FireMarshal itself. Any Qemu-only deps are a historical artifact and I'm fine removing them completely.

PR #121, opened 2020-01-23
Expand quickstart to a broader 'tutorials' section and clean up a bunch

NathanTP:
This cleans up a bunch of stuff in the quickstart/tutorial section of the docs. Especially since we're removing even more Marshal docs from firesim, I wanted to make sure everything was up to date.

### DIFF
--- a/docs/source/Tutorials/quickstart.rst
+++ b/docs/source/Tutorials/quickstart.rst
@@ -1,5 +1,7 @@
-Quick Start
---------------------------------------
+.. _tutorial-quickstart:
+
+Quick Start (Built-In Workloads)
+=========================================
 
 After a fresh clone of the FireMarshal repository, you may need to update
 submodules. You may skip this step if you only intend to build bare-metal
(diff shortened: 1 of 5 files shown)


docs/source/Tutorials/quickstart.rst
```diff
@@ -59,8 +69,8 @@ you must first install it from FireMarshal:
 
 ::
 
-  ./marshal install workloads/br-base.json
+  ./marshal install br-base.json
 
 This command creates a firesim workload file at
 ``firesim/deploy/workloads/br-base.json``. You can now run this workload using
-the standard FireSim commands. 
+the standard FireSim commands.
```
sagark: Can we add this back in here? (from what used to be in the firesim docs)

Simply change the
``workloadname`` option in ``firesim/deploy/config_runtime.ini`` to
"fedora-uniform.json" and then follow the standard FireSim procedure for
booting a workload (e.g. :ref:`single-node-sim` or :ref:`cluster-sim`).

NathanTP: Fixed in the FireSim docs instead. See https://github.com/firesim/firesim/pull/465 (in Booting-Fedora.rst)

PR #129, opened 2020-02-10
FireMarshal changes for TraceRV + Stack Unwinding + Flame Graph support from FirePerf

sagark:
See bullet points 2 and 3 on https://github.com/firesim/firesim/pull/496


sagark:
@a0u @NathanTP @alonamid this is ready for review.

PR #141, opened 2020-04-14
[RFC] Add post binary script run option

abejgonzalez:
Adds a `post-bin` option to the workload configuration so that you can run a script after the kernel was built (but before the image is created). This should be the proper way to build external Linux modules since you need to first build the kernel, then build the external module, then you can insert the module during boot (using the `init.d` directory).

This change is needed for the NVDLA driver since it requires you to build the kernel with a specific `kconfig` before building the driver itself. Thus, using `host-init` is not an option (even if you do the proper steps to get a correct `.config` file in the `riscv-linux` directory i.e. `merge_config.sh` and `make modules_prepare`)

*TODO:*
- [x] Unit test
- [x] Determine how to deal with multiple jobs (double-check dep. tree)
- [x] Review / Cleanup documentation


docs/source/workloadConfig.rst
```diff
@@ -80,6 +80,17 @@ explicitly build this workload. This option may include arguments for the script
 However, any affects that host-init has on the resulting rootfs *will* be
 reflected in the child.
 
+bin-post
+--------------
+A script to run natively on your host (i.e., them machine where you
+invoked FireMarshal) from the workload source directory after you build the
+binary(s) for this workload but before the image(s) is/are built. This option may
+include arguments for the script, e.g. ``"bin-post" : "foo.sh bar baz"``.
+
```
NathanTP, line 89: it might be nice to include some examples of why you would do this. It took some thinking on our part to figure out that we needed this for building drivers. Perhaps later we can include a more detailed example of including custom kernel modules (not needed for this PR). Some reasons for post-bin:
* Building kernel modules against the specific kernel and kconfig that are being used.
* Analyzing the kernel binary to extract debugging or tracing information

It might be good to also document what the user can expect at this phase:
* The binary will be built, but the image will not
* The kernel configuration used to build the kernel will be at FIREMARSHAL_LINUX_SRC/.config
* Any build artifacts generated by building the kernel will continue to be present (we will not run 'make clean' or equivalent in the kernel source).

docs/source/workloadConfig.rst
```diff
@@ -80,6 +80,17 @@ explicitly build this workload. This option may include arguments for the script
 However, any affects that host-init has on the resulting rootfs *will* be
 reflected in the child.
 
+bin-post
```
NathanTP, line 83: Could we call this 'post-bin'? If I'm talking about it I'd say 'this script runs post-bin' not 'this script runs bin-post'.

test/bin-post2.json
```diff
@@ -0,0 +1,21 @@
+{
```
NathanTP, line 1: It's not required, but you can use the same workdir for both post-bin.json and post-bin2.json if you want to make it a bit DRYer. All you have to do is set the '--workdir' option in post-bin2.json (or both of them).

wlutil/build.py
```diff
@@ -389,7 +413,6 @@ def makeImage(config):
         # Apply and run the init script
         init_overlay = config['builder'].generateBootScriptOverlay(str(config['guest-init'].path), config['guest-init'].args)
         applyOverlay(config['img'], init_overlay)
-        print("Launching: " + str(config['bin']))
```
NathanTP, line 392: Good catch. This used to be needed because I couldn't figure out how to supress the output of Qemu when running the guest-init and it was confusing to users. Since I fixed it, this isn't needed. Thanks.

wlutil/build.py
```diff
@@ -43,7 +43,16 @@ def handleHostInit(config):
            raise ValueError("host-init script " + str(config['host-init']) + " not found.")
 
        run([config['host-init'].path] + config['host-init'].args, cwd=config['workdir'])
- 
+
+def handleBinPost(config):
```
NathanTP, line 47: Since marshal can be run from anywhere these days, and kernel sources can move around and stuff, I think it'd be a good idea to tell the post-bin script where the kernel binary and source are. At a minimum, it will make the scripts easier to write, but it will also make the system less brittle. I'm thinking we could add two environment variables for the script to use:
FIREMARSHAL_LINUX_SRC
FIREMARSHAL_LINUX_BIN

The SRC one will point the source actually used to build the bin which could be the default one, or in a board directory, or in the workload's workdir directory (it's whatever is in config['linux-src']).

we'd also need to add these to the docs.

wlutil/build.py
```diff
@@ -184,10 +210,9 @@ def buildWorkload(cfgName, cfgs, buildBin=True, buildImg=True):
     global taskLoader
     if taskLoader == None:
         taskLoader = buildDepGraph(cfgs)
-        
+
     config = cfgs[cfgName]
 
-    # handleHostInit(config)
```
NathanTP, line 190: Nice catch!

wlutil/build.py
abejgonzalez: Oh I think this was a mistake on my end. I had other print statements that I was using that I wanted to delete so I did a blanket delete. At least this is still good!

abejgonzalez:
Ok tried to address the comments. LMK if you see something else that needs to change.

wlutil/build.py
```diff
@@ -44,14 +44,19 @@ def handleHostInit(config):
 
        run([config['host-init'].path] + config['host-init'].args, cwd=config['workdir'])
 
-def handleBinPost(config):
+def handlePostBin(config, linuxSrc, linuxBin):
```
NathanTP, line 47: linuxSrc is config['linux-src'] and linuxBin is config['bin']. I can't think of a scenario where that wouldn't be true so you don't really need separate args for it.

abejgonzalez: I think the `linux-src` is fine to get rid from the arg list. However, `linux-bin` can either be a "no-disk" or a "normal" version so just using `config['bin']` is not sufficient (unless Im mistaken). So when I invoke the function I give it the direct path to the "right" binary i.e. the "-nodisk" or the normal path.

NathanTP: alright, that's fine.

PR #142, opened 2020-04-15
Bump Linux

abejgonzalez:
Bumps Linux. See https://github.com/firesim/linux/pull/1


NathanTP:
This is really old. If we still need to bump Linux, let's do it in a new PR. I tried a little while back and had some issues so it's not trivial. closing.

PR #143, opened 2020-04-16
Full support for workload kfrags

NathanTP:
Workload kernel config fragments now inherit from parent configs. This also introduces a test to make sure that kfrags actually work.


NathanTP:
It turns out we were just getting lucky with the GDB workload. I never tried it on FireSim (since it wouldn't work anyway). Qemu will boot OK with the kernel defconfig, it doesn't actually require anything from br-base's kconfig. So what was happening is that the gdb kfrag was being applied on top of defconfig and failed to include anything from br-base. This PR makes it so that the entire chain of parents get merged.

Issue #144, opened 2020-04-21
Workload Launch Arguments


NathanTP:
It would be nice to be able to pass launch-time arguments to the workload. This would allow users to build a single large workload with many tests, and then just tell it which test to run at launch time instead of building a bunch of independent workloads (and their associated disk cost).

This might be possible with the kernel command line, or a quick modification of the base image right before launch. Whatever we do should work in FireSim.

NathanTP:
Closing due to lack of interest. This doesn't seem to be a pressing need for anyone. Plus, re-using the same image for multiple tests could get funny if the test modifies the image or if we want to launch them in parallel.

Issue #134, opened 2020-02-17
workdir search paths


NathanTP:
The search paths for bases are somewhat messy and static right now. There is the --workdir option (which has a confusing name btw) and some builtin defaults (e.g. boards/firechip/default-workloads), but these are disparate mechanisms. Now that we have a proper configuration system, we should add a FIREMARSHAL_PATH option (or whatever we call it) that's like PATH in your shell but for workloads. If you set it, you can simply run 'marshal build workload.json' and if workload.json is in any of those paths, it will build. We could still override/append with --workdir or config files.

NathanTP:
This has been addressed. The option is MARSHAL_WORKLOAD_DIRS or 'workload-dirs' in your marshal-config.yaml. Not sure when that got added.

PR #146, opened 2020-04-21
Example workload for running multiple tests over the same base

NathanTP:
A pain point (brought up in the mailing list) is that if you want to run a bunch of variations of a test on the same base image, you'll end up with a million copies of the image. Instead of creating a bunch of different workloads, you can trick Marshal into modifying the same image over and over.

It's not pretty because of #134 and #144. Once those are resolved, this should get a lot cleaner. For now though, I think it's worth documenting the workaround.

This PR only provides an additional example, it does not modify marshal at all.


example-workloads/many-tests/README.md
```diff
@@ -0,0 +1,44 @@
+# Supporting Many Tests On One Base
+In same cases, you would like to run many different tests over the same image.
+If you were to make N different workloads, one for each test, you'd end up with
+N coppies of the binary and image, along with the associated build time of
```
a0u, line 4:
```suggestion
N copies of the binary and image, along with the associated build time of
```

example-workloads/many-tests/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Basic Strategy
+The basic strategy is as follows:
+* Create a single base workload that contains everything needed to run every test.
+* Create many copies of a test workload, one for each test you want to run:
+    * All test json's should have the same workload name (the "name" option).
+    * Each test json should be in its own subdirectory (otherwise FireMarshal
+      will try to load them all and complain that multiple workloads have the
+      same name)
+    * Each test workload should have it's own workdir (the subdirectory), it's
+      own 'run' or 'command' option, and optionally it's own testing directory.
```
a0u, line 17:
```suggestion
    * Each test workload should have its own workdir (the subdirectory), its
      own 'run' or 'command' option, and optionally its own testing directory.
```

example-workloads/many-tests/README.md
```diff
@@ -0,0 +1,44 @@
...
+FireMarshal just thinks you modified the workload json and rebuilds only the
+changed parts.
+
+## Running the Example
+To use this example, you can run the 'runTest.sh' script:
+
+    $ ./runTest.sh t0/t.json
+    $ ./runTest.sh t1/t.json
+
+This script coppies the test workload into the top-level directory and runs
```
a0u, line 33:
```suggestion
This script copies the test workload into the top-level directory and runs
```

NathanTP:
Fixed. Ready for re-approval.

PR #148, opened 2020-04-30
Disable the SiFive UART driver until we can fix its baud rate

davidbiancolin:
In the most recent RC bump the Sifive UART driver started working on firechip -- the difficulty now is that it is being configured to use a different baud rate (at multiple different points during boot) than the one our simulation bridge expects. This breaks console output after the reconfiguration. 

Until we find a way to fix the baud rate to our expected value in all of the places it is being set continue relying on the SBI serial driver. 

(See firesim/firesim#540)


davidbiancolin:
> Does `CONFIG_SIFIVE_PLIC` remain enabled when `CONFIG_SOC_SIFIVE` is unselected?

No, only the two settings i changed are changed in the resulting config. I've now checked both the BR and fedora configs.

> The same should also be applied to the Fedora config. (I noticed that one still uses the obsolete `CONFIG_RISCV_PLIC` option. We should probably regenerate the defconfig another time to keep up with the churn upstream.)

I've updated the fedora config. I'm deferring the upgrade to a modern defconfig because apparently I still have other RC-bump fires to put out.

PR #152, opened 2020-05-15
Opensbi

NathanTP:
Switches to OpenSBI by default, still (sort of) supports bbl. It seemed as good a time as ever so I refactored some of the SW dependencies into the boards directory to clean the top-level up a bit.

This is a draft PR for a few reasons:

1. We will require updated versions of qemu and spike so this PR will depend on an upcoming change in chipyard (see the bumpSpikeQemu branch in chipyard).
1. Binaries built with bbl don't poweroff in qemu. This is doesn't have anything directly to do with the opensbi changes, it seems to just be an incompatibility with the updated Qemu. I'd appreciate it if someone with more experience there could look into that.
1. It has not been tested in FireSim. Can someone do that?


NathanTP:
Some "updates":
1. I don't know why BBL-based linux workloads hang on poweroff on qemu. I suggest we simply 'deprecate' bbl and only enable for specific things that only run on spike (at least until we fix the problem, if ever).
2. There does not appear to be any uart output when running openSBI workloads on firesim. It works fine on spike, but there's no output after 'commencing simulation' in firesim. Someone who understands that crap wanna take a look (@sagark)?

The best way to test is using the [OpenSBI branch of chipyard](https://github.com/ucb-bar/chipyard/tree/opensbi). Although the only chipyard changes are bumps to spike and qemu.

NathanTP:
one other tidbit that might be related:
OpenSBI seems to use "\r\r\n" as a newline while BBL uses "\r\n". It's the same between qemu and spike and it's been breaking all my tests. It could be related to the firesim issue. Any insights someone can share would be helpful.

a0u:
Most likely the UART divisor isn't being set properly for the higher baud rate that the FireSim bridge expects.   A perusal of the OpenSBI code suggests that the baud rate and input clock frequency are being taken from the device tree: [opensbi/lib/utils/serial/fdt_serial_sifive.c](https://github.com/riscv/opensbi/blob/4f18c6e55049d858c62e87d2605dd41c06956e4e/lib/utils/serial/fdt_serial_sifive.c).
Also make sure that the platform-specific config of OpenSBI actually includes the serial driver.

If it works, the FDT option seems cleaner instead of hard-coding the divisor as we did with the pk fork.

a0u:
It appears that the QEMU `virt` board emulates the SiFive test device for signalling poweroff; HTIF is not supported.  There was a recent change that added the DT entries for it to work with the generic `syscon-poweroff` kernel driver: qemu/qemu@0e404da00737eaa69204e7f3446b85b57d397e4a.

As of Linux 5.7, selecting the `SOC_VIRT` option enables:
```
select POWER_RESET_SYSCON
select POWER_RESET_SYSCON_POWEROFF
```

NathanTP:
For the uart thing: I'm currently building the "generic" platform. The only mention of serial in that platform.c is the fdt_serial thing. The fu540 platform uses sbi_utils/serial/sifive-uart.h. I don't know the difference or how it relates to our SoCs. As for setting the baud thing, I was under the impression that the SoC creates the device tree, so if opensbi really is using the baud from the dt then it ought to be doing the right thing right?

For poweroff: We are already selecting SOC_VIRT so POWER_RESET_SYSCON is enabled in the kernel. We're using QEmu 5.0.0 which is more recent than the cited commit. Also, this is only a problem on bbl. OpenSBI shuts down just fine which would indicate that the kernel is trying to do the right thing but not playing nice with something in BBL. Maybe BBL messes with the device tree somewhere?

a0u:
FireSim uses a non-standard baud rate of 3686400.  The rocket-chip-generated DTS omits the `current-speed` property, so OpenSBI defaults to `DEFAULT_SIFIVE_UART_BAUD`, defined to be 115200.

https://github.com/riscv/opensbi/blob/4f18c6e55049d858c62e87d2605dd41c06956e4e/lib/utils/fdt/fdt_helper.c#L192

a0u:
Regarding poweroff, the issue is that the Kconfig has `CONFIG_RISCV_SBI=y`, so the `pm_power_off` handler is set to the [SBI implementation](https://github.com/firesim/linux/blob/88622707f72ef05d7768e4165ab91cd892fd9d97/arch/riscv/kernel/sbi.c#L552).  I suspect that overrides the `syscon-poweroff` driver - check the kernel logs for any messages like "pm_power_off already claimed".

OpenSBI works because it supports the [SiFive test device](https://github.com/riscv/opensbi/blob/6585fabbcca24bf1c1df0582d048d93b3caf7ce7/lib/utils/sys/sifive_test.c) through the legacy SBI call, unlike bbl.  This binds to the `sifive,test1` DT node presented by the QEMU `virt` board.

a0u:
@NathanTP: Upstream riscv-pk now has a fix for the poweroff issue: riscv/riscv-pk@5d9ed238e1cabfbca3c47f50d32894ce94bfc304.  Contrary to my previous comment, bbl already had support for the SiFive test finisher device but failed to match on the `compatible` property exposed by QEMU.

Previously, the `virt` board in qemu/qemu@4f59102571fce49af180cfc6d4cdd2b5df7bdb14 (to which Chipyard currently points) had set `compatible` to `"sifive,test0"`.  However, QEMU master now uses `"sifive,test1\0sifive,test0\0syscon"`.  While this change is necessary for OpenSBI, which requires `sifive,test1`, it so happened that bbl could not parse the multi-value format correctly.

NathanTP:
Great, this fixed the bbl+qemu issue. Now we just have the baud rate issue on FireSim.

a0u:
Based on the lack of TracerV progress, a more immediate problem seems to be that the DTB copy loop, which precedes any UART interaction, is hanging on a memory access somewhere:

https://github.com/riscv/opensbi/blob/518e85cccb18a6dfe71fc8d1227c8f8d74527dc9/firmware/fw_base.S#L337-L342

a0u:
When copying the FDT from bootROM to DRAM, OpenSBI determines the FDT size incorrectly when the source FDT is not XLEN-aligned.  The testchipip bootROM places the DTB at 0x1009c.

OpenSBI aligns the pointer:
https://github.com/riscv/opensbi/blob/518e85cccb18a6dfe71fc8d1227c8f8d74527dc9/firmware/fw_base.S#L303-L305

...but does not adjust the offset to the `totalsize` field in the FDT header:
https://github.com/riscv/opensbi/blob/518e85cccb18a6dfe71fc8d1227c8f8d74527dc9/firmware/fw_base.S#L308

Patch to be sent upstream:
```diff
diff --git a/firmware/fw_base.S b/firmware/fw_base.S
index b66ac41..6a7a143 100644
--- a/firmware/fw_base.S
+++ b/firmware/fw_base.S
@@ -305,9 +305,9 @@ _scratch_init:
        and     t0, t0, a3
        /* t2 = source FDT size in big-endian */
 #if __riscv_xlen == 64
-       lwu     t2, 4(t0)
+       lwu     t2, 4(a1)
 #else
-       lw      t2, 4(t0)
+       lw      t2, 4(a1)
 #endif
        /* t3 = bit[15:8] of FDT size */
        add     t3, t2, zero
@@ -332,6 +332,9 @@ _scratch_init:
        or      t2, t2, t5
        /* t2 = destination FDT end address */
        add     t2, t1, t2
+       /* Adjust end address if source FDT is not __SIZEOF_POINTER__-aligned */
+       sub     t3, a1, t0
+       add     t2, t2, t3
        /* FDT copy loop */
        ble     t2, t1, _fdt_reloc_done
 _fdt_reloc_again:
```

Unfortunately, I am hitting another deadlock after this.

a0u:
I have UART working for OpenSBI on FireSim.  To summarize the prerequisites:

* Align DTB to 8-byte boundary: ucb-bar/testchipip#90
* Emit `htif` node in device tree: ucb-bar/chipyard#607 (needed to register the correct reset handler)
* Bump riscv-pk: [dev-bump-20200618](/firesim/riscv-pk/tree/dev-bump-20200618) (also eliminate the hardcoded `htif = 1` hack with the above)
* Bump QEMU to 5.0.0 in Chipyard

Setting the correct UART divisor in OpenSBI can be done with a crude hack that parallels bbl:
```diff
diff --git a/lib/utils/fdt/fdt_helper.c b/lib/utils/fdt/fdt_helper.c
index 887d6ed..f41074e 100644
--- a/lib/utils/fdt/fdt_helper.c
+++ b/lib/utils/fdt/fdt_helper.c
@@ -21,8 +21,8 @@
 #define DEFAULT_UART_REG_SHIFT         0
 #define DEFAULT_UART_REG_IO_WIDTH      1
 
-#define DEFAULT_SIFIVE_UART_FREQ               0
-#define DEFAULT_SIFIVE_UART_BAUD               115200
+#define DEFAULT_SIFIVE_UART_FREQ               3200000000
+#define DEFAULT_SIFIVE_UART_BAUD               3686400
 #define DEFAULT_SIFIVE_UART_REG_SHIFT          0
 #define DEFAULT_SIFIVE_UART_REG_IO_WIDTH       4
```

Ideally, the `clock-frequency` and `current-speed` properties would be set in the FDT to avoid hardcoding these values.  That might be possible with a sifive-blocks wrapper module.  We can also imagine extending the simulation driver to rewrite the generated DTB and place it in memory during LOADMEM.

Issue #157, opened 2020-06-16
Generated Overlays


NathanTP:
If the contents of an overlay are generated by a host init, they will not be detected. This is because Marshal determines the list of files at the time of config loading (early on). Since overlays are globbed, we won't see the files yet. This is not an issue for 'files' since they are explicitly listed and marshal won't even look for them until after host-init.

A simple-ish fix is to do the glob in the image build task (that runs after host-init), but this would not add the generated files to the list of dependencies (marshal would not rebuild if they change). While this probably isn't an issue most of the time (they were generated anyway) it could lead to some very confusing issues if you do modify them.

A more robust solution would be to add the dependencies after host-init, but this could be challenging in doit (I'm not even sure if it's possible to dynamically edit the dependency tree during execution).

NathanTP:
@a0u

NathanTP:
Resolved by #158

PR #158, opened 2020-06-16
Generated overlay

NathanTP:
Addresses #157 

File and Overlay handling used to happen at config load time which could miss generated inputs (e.g. from host-init). This checking has now been moved to runtime and ordered right before image creation (after anything might have been generated). It also cleans up some of the file handling in general.

### DIFF
--- a/wlutil/build.py
+++ b/wlutil/build.py
@@ -66,7 +66,7 @@ def fileDepsTask(name, taskDeps=None, overlay=None, files=None):
     taskDeps should be a list of names of tasks that must run before
     calculating dependencies (e.g. host-init)"""
 
-    def overlayDeps(overlay, files):
+    def fileDeps(overlay, files):
         """The python-action for the filedeps task, returns a dictionary of dependencies"""
         deps = []
         if overlay is not None:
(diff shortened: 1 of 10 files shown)


wlutil/build.py
```diff
@@ -58,6 +58,38 @@ def handlePostBin(config, linuxBin):
...
+
+        for dep in deps.copy():
+            if dep.is_dir():
+                deps += [ child for child in dep.glob('**/*') ]
+
+        return { 'file_dep' : [ str(f) for f in deps if not f.is_dir() ] }
+
+    task = {
+            'name' : 'calc_' + name + '_dep',
+            'actions' : [ (fileDeps, [overlay, files]) ],
```
a0u, line 86: If I understand correctly, `fileDeps(overlay, files)` would be called after `img_task_deps` (i.e., `host-init`) and the results dynamically injected into the task graph?  It seems that this resolves #157 without the anticipated limitations.

NathanTP: ya, what this is doing is moving the dependency checking to its own task instead of doing it when building the initial dependency graph. That task outputs a list of dynamically generated dependencies (that's what fileDeps does). The nice thing here is that it's a task like any other and I can give it dependencies of its own (host-init in this case).

Issue #131, opened 2020-02-13
Changes to board drivers not detected


NathanTP:
Marshal does not seem to register the board drivers as dependencies for workloads, so updates to the driver don't trigger a rebuild.

NathanTP:
Resolved by #165

PR #165, opened 2020-09-04
User drivers

NathanTP:
The primary change here is the ability to specify kernel modules to build and load automatically. This serves two purposes:

1. Removes some of the complexity of building kernel modules (a somewhat tricky task) and especially loading them automatically. This is particularly important if you need the module to load early for a boot-critical device. You can still do it in a post-bin script if you want, this is just easier.
1. Allows users to override default system drivers. This is particularly important for when you specify a custom kernel that doesn't have the exact same version as the default (drivers aren't generally portable between kernel releases).

This PR also puts all the linux-related options in their own config section instead of spread around the top-level config. The old 'linux-src' and 'linux-config' options still work but are deprecated.


NathanTP:
This also fixes #131

NathanTP:
FYI: @abejgonzalez @dayeol

a0u (approved):

abejgonzalez:
Looks nice. When this gets merged into Chipyard, Ill try to port the NVDLA workload and driver to this format.

PR #166, opened 2020-09-17
Inherit Parent's Binary

NathanTP:
Currently, marshal is conservative and builds every workload's binary from scratch. When you have a bunch of workloads that don't change anything related to the Linux build, this becomes a big performance issue. Linux compilation is the slowest step in the build process anyway, repeating it unnecessarily is particularly painful.

This PR makes it so that Marshal just copies the parent's binary if the child workload wouldn't have changed anything in it. Users with lots of workloads should see a big performance boost.

The risk here is missing corner cases, I think I found everything that might affect the output binary but there's always a chance that I missed something. The current criteria are:

- Different 'linux' option (this was added recently and groups all the linux-related config together)
- Different 'firmware' option
- Different 'host-init' - you can do anything in host-init so we have to assume you might do something that affects the binary.
- all no-disk builds are done from scratch


wlutil/config.py
```diff
@@ -411,6 +419,20 @@ def applyBase(self, baseCfg):
             self.cfg['run'] = getOpt('wlutil-dir') / 'null_run.sh'
             self.cfg['runSpec'] = RunSpec(script=self.cfg['run'])
 
+        # To avoid needlessly recompiling kernels, we check if the child has
+        # the exact same binary-related configuration.
+        self.cfg['use-parent-bin'] = True
```
a0u, line 424: Setting `"use-parent-bin": false` should unconditionally force a kernel rebuild?

wlutil/config.py
```diff
@@ -411,6 +419,20 @@ def applyBase(self, baseCfg):
...
+        for opt in ['firmware-src', 'host-init']:
+            if opt in self.cfg and self.cfg.get(opt, None) != baseCfg.get(opt, None):
+                self.cfg['use-parent-bin'] = False
+
+        if 'linux' in self.cfg and 'linux' in baseCfg:
+            for opt in self.cfg['linux'].keys():
+                if self.cfg['linux'].get(opt, None) != baseCfg['linux'].get(opt, None):
```
a0u, line 431: What if both `linux` options have identically named configs/files but relative to different workdirs?  Are the paths fully resolved when comparing them?

NathanTP: All paths are converted to absolute early in the config loading process. For the linux stuff this happens at line 285 and 287.

wlutil/config.py
NathanTP: not unconditionally. "use-parent-bin" = False means that if the child binary isn't up to date, the whole kernel should be rebuilt (if nothing has changed, it won't be rebuilt). "use-parent-bin" = True means that if the child binary isn't up to date, don't compile a whole new binary, just copy your parent's.

In other words, that flag only affects what it means to generate a binary for a workload (copy from the parent vs compile from scratch).

a0u (approved):

PR #167, opened 2020-09-22
Fix kernel module dependency checking with host-init

NathanTP:
Kernel module dependency checking could not handle the case where the
    module was created dynamically in a host-init script. Module status
    checking now has no extra requirements. Instead it will issue a warning
    if it cannot determine the status of a kernel module and will assume the
    binary is not up to date. If there is really a problem with the module,
    it will show up when actually attempting to build the module.

There is an alternative approach: make kmod dependency checking depend on host-init. In this case, host-init would have to run before we can even check to see if the rest needs to be rebuilt. In theory, this may lead to the host-init running when it didn't have to, but in reality they seem to run anyway (I'm not totally sure why that is, something else probably depends on it). I prefer the approach done here because it decouples components more (I would want issues with the host-init to only show up when host-init was supposed to run, not because we were checking dependencies). Either way would be fine.

### DIFF
--- a/wlutil/build.py
+++ b/wlutil/build.py
@@ -73,23 +73,38 @@ def submoduleDepsTask(submodules, name=""):
             }
 
 
-def kmodDepsTask(cfg, name=""):
+def kmodDepsTask(cfg, taskDeps=None, name=""):
     """Check if the kernel modules in cfg are uptodate (suitable for doit's calc_dep function)"""
-    def checkMods(cfg):
-        for driverDir in cfg['linux']['modules'].values():
-            p = run(["make", "-q", "LINUXSRC=" + str(cfg['linux']['source'])], cwd=driverDir, check=False)
 
-            if p.returncode != 0:
+    def checkMods(cfg):
+        log = logging.getLogger()
+        for driverDir in cfg['linux']['modules'].values():
+            if not driverDir.exists():
+                log.warn("WARNING: Required module " + str(driverDir) + " does not exist: Assuming the workload is not uptodate.")
                 return False
+            try:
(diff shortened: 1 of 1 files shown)


NathanTP:
FYI: @dayeol

wlutil/build.py
```diff
@@ -73,24 +73,39 @@ def submoduleDeps(submodules):
...
+
     def checkMods(cfg):
+        log = logging.getLogger()
         for driverDir in cfg['linux']['modules'].values():
-            p = run(["make", "-q", "LINUXSRC=" + str(cfg['linux']['source'])], cwd=driverDir, check=False)
+            if not driverDir.exists():
+                log.warn("WARNING: Required module " + str(driverDir) + " does not exist: Assuming the workload is not uptodate.")
+                return False
+            try:
+                p = run(["make", "-q", "LINUXSRC=" + str(cfg['linux']['source'])], cwd=driverDir, check=False)
```
a0u, line 86: Shouldn't the subprocess already raise an error if `cwd` doesn't exist?

NathanTP: Ya, it did raise an exception that wasn't being caught. I could add it to the except if I wanted, might be a bit cleaner looking. I do want to keep the two cases separate (filenotfound vs other) just for clarity to the user. I can switch it to do both in the except.

a0u: It's fine to me as it is.  `FileNotFoundError` is a bit more general than `driverDir` not existing, for example if `make` is not found.

a0u (approved):

a0u (approved):

PR #172, opened 2020-10-01
Firmware Args

NathanTP:
Add the 'opensbi-build-args' and 'bbl-build-args' options for additional firmware configuration. This also documents and properly tests the existing 'bbl-src' and 'opensbi-src' options that had some issues.

This feature was requested by the keystone team.


docs/source/workloadConfig.rst
```diff
@@ -120,16 +120,34 @@ The specified module directory must contain a Makefile that can be invoked as:
...
+
+bbl-build-args
+^^^^^^^^^^^^^^^^^^^^
+Additional arguments to be passed to the configure script for bbl. Arguments
+will be appended to any system defaults.
 
 use-bbl
 ^^^^^^^^^^^^^
 Use BBL as the firmware rather than the default OpenSBI. If you specified
 'pk-src' (pk includes bbl), that will be used, otherwise the default
```
a0u, line 149:
```suggestion
'bbl-src', that will be used, otherwise the default
```

a0u (commented):
Given that #165 reorganized the `linux` options into their own hierarchy, should we not use a similar scheme such as `opensbi/source` and `opensbi/build-args` for consistency?

NathanTP:
I agree on the firmware option group thing. I pushed some commits for that. It got a little out of hand and turned into a bit of a refactor of config.py. Sorry lol. It should be more readable and maintainable now though.

One thing to note is that it breaks backwards compatibility with the openSBI PR from a little while ago (since some firmware options were introduced there). Shouldn't be a problem since it only affects dev (doesn't break anything in 1.9). @dayeol may need to update some keystone configs though.

NathanTP:
@dayeol please test this and confirm it meets keystone's needs.

test/bbl/copy-src.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+if [ ! -d linux ]; then
+  rsync --exclude ".git" -r ../../boards/firechip/firmware/riscv-pk .
```
a0u, line 5: It doesn't matter much, but here's an alternative to `rsync` that ignores uncommitted changes in the working copy:
```sh
git -C ../../boards/firechip/firmware/riscv-pk archive --format=tar --prefix=riscv-pk/ HEAD | tar xf -
```

a0u (dismissed):
Yeah, none of the overhauled options were part of the last release, so it shouldn't break any stable workloads.

(Approved pending successful testing by @dayeol.)

dayeol:
Just tested it, it works great. I appreciate it!

dayeol (approved):

PR #174, opened 2020-10-13
Launch multi jobs

NathanTP:
Add support for launching multiple jobs from a single "launch" command. This is pretty handy for testing multi-job workloads, especially if they have a post-run-hook. Before this, there wasn't an easy way to test multi-job post-run-hooks (FireSim could do it, but marshal couldn't). Since we don't have a network model, we can't really run multi-job workloads. Instead we just run them sequentially. This lays the groundwork for future improvements though.


marshal
```diff
@@ -39,8 +39,8 @@ def main():
         'launch', help='Launch an image on a software simulator (defaults to qemu)')
     launch_parser.add_argument('-s', '--spike', action='store_true',
             help="Use the spike isa simulator instead of qemu")
-    launch_parser.add_argument('-j', '--job', nargs='?', default='all',
-            help="Launch the specified job. Defaults to running the base image.")
+    launch_parser.add_argument('-j', '--job', nargs='?', default=None, type=lambda s: [item for item in s.split(',')],
```
a0u, line 42: It seems unlikely for a job name to intentionally include a comma, but if so, we could instead allow multiple `-j` options to accumulate  (e.g., `-j foo -j bar -j baz`):

```C
launch_parser.add_argument('-j', '--job', action='append')
```

a0u (approved):

marshal
NathanTP: Ya, that's fair. It's also more clear that way, the list was somewhat of a hack and not really a standard thing. I've switched it to multiple -j options instead of a list. This required adding a new '--all' argument to specify launching all jobs.

a0u (dismissed):
Looks good to me.

NathanTP:
@a0u: I needed to merge in some of the other PRs and fix a test. Can you re-approve?

a0u (approved):

PR #175, opened 2020-10-27
Bump Fedora

alonamid:
Bump the Fedora image to a more recent image (Fedora 31).
This requires some additional manipulation of the downloaded image since it also includes a boot partitions. That's the reason fstab is added to the overlay.
Tested on FireSim and using the marshal tests script.


wlutil/fedora/overlay/etc/fstab
```diff
@@ -0,0 +1 @@
+UUID=12bd5ff0-295f-40fe-a8ba-d08878f38173  / ext4    defaults,noatime,x-systemd.device-timeout=300s,x-systemd.mount-timeout=300s 0 0
```
a0u, line 1: (For the record, a patched `/etc/fstab` is necessary since the marshal images omit the boot partition.)

While the `fstab` entry for `/` is mostly irrelevant since the initramfs is responsible for mounting the rootfs in our current setup, the hardcoded UUID could become an unwelcome surprise if that ever changes in a future Fedora bump.

Unfortunately, an explicit device node won't generalize to both QEMU (`/dev/vda`) and FireSim (`/dev/iceblk`), but forunately, the base filesystem comes with a default label that can be used:
```suggestion
LABEL=_/  / ext4    defaults,noatime,x-systemd.device-timeout=300s,x-systemd.mount-timeout=300s 0 0
```

a0u (commented):
@NathanTP: Are there any tests that depend on Fedora boot messages which also require updating?

NathanTP:
None of the tests should rely on Fedora boot messages because we strip most of that out before comparing. There's lots of non-deterministic output anyway.

boards/firechip/base-workloads/fedora-base/overlay/etc/yum.repos.d/fedora-riscv.repo
```diff
@@ -1,21 +1,17 @@
 [fedora-riscv]
 name=Fedora RISC-V
-baseurl=http://fedora-riscv.tranquillity.se/repos-dist/f29/latest/riscv64/ https://dl.fedoraproject.org/pub/alt/risc-v/repo/fedora/29/latest/riscv64/ https://mirror.math.princeton.edu/pub/alt/risc-v/repo/fedora/29/latest/riscv64/
+baseurl=http://fedora.riscv.rocks/repos-dist/rawhide/latest/riscv64/
```
NathanTP, line 3: The backups were originally added because the main repo kept going down. I guess the official fedora repo is probably reliable enough. Are there other mirrors we could add just in case?

NathanTP (commented):
Assuming scripts/fullTest.py passes, and something reasonable boots on FireSim, I'm happy.

boards/firechip/base-workloads/fedora-base/overlay/etc/yum.repos.d/fedora-riscv.repo
a0u: I'm not aware of any.  fedora.riscv.rocks is used by Red Hat for nightly builds.  The "main" repo (https://dl.fedoraproject.org/pub/alt/risc-v/repo/fedora/rawhide/latest/riscv64/) hasn't been updated since mid-2019.

alonamid:
Tested again on both FireSim and QEMU after commiting Albert's suggestions. This should be ready

a0u (approved):

NathanTP (approved):
looks good. looking forward to a less dodgy fedora.

PR #177, opened 2020-10-27
Disable Inherited Modules

NathanTP:
modules can now have a null path to disable an inherited module. This is needed in order to disable default modules like device drivers.

draft, pending tests


NathanTP:
tests pass, ready for review

a0u (approved):

PR #180, opened 2020-11-18
bump icenet driver for checksum fix

zhemao:
See https://github.com/firesim/icenet-driver/pull/3 for discussion.


NathanTP:
This looks fine in functional sim, but that doesn't really use the driver much (just verifies that it loads, not that it works). Marshal doesn't have a good NW model or a functional model of the NIC so it's hard to test locally. If you can test on FireSim and confirm it works there, I'll approve.

NathanTP (approved):
@zhemao confirms this works on FireSim (it requires the head of chipyard dev to work though). There shouldn't be any risks to marshal, but we should remember to document the new HW requirement in the release notes.

Issue #43, opened 2019-07-10
Optionally re-build from changed dependencies


NathanTP:
In an abundance of caution, marshal currently does not inherit changes from dependencies if you don't clean first. This is to avoid wiping out ephemeral data on your rootfs accidentally. It's not a big deal to just "marshal clean work.json && marshal build work.json" if you want the new dependencies, but a "marshal build --force" might be more clear. This would re-inherit dependencies all the way down the chain if needed, but will remove any ephemeral data you had laying around in your rootfs.

NathanTP:
thinking about this some more, I think there should be a warning or "type yes to continue" type of thing if a parent changed.

NathanTP:
After more discussion there seem to be a few options on how to address this:

1. Add an interactive warning if marshal is going to overwrite your workload. e.g. "Parent workloads have changed. Overwrite your workload image with the new parent (all local changes will be lost)? Y/N"
    1. May lead to annoying behavior when you try to 'fire and forget', it makes 'marshal build' potentially interactive and could break scripts.
2. Change the default behavior to just always overwrite (how common is it for people to have important manual changes in their workload?). I'm not sure exactly how Docker handles this.
    1. Could cause some annoying surprises from time to time (if you were in the middle of manually tweaking the workload). Honestly I don't think this is very common anymore, everything should be in your workload config and it's pretty fast to rebuild these days.
3. Keep the current default behavior, but add a "--force" flag that will overwrite the child on parent changes. There would also be a new warning that parent changes were not applied and to use '--force' if you're willing to lose local changes. This seems similar to other tools like git.
    1. Is probably the safest, but I suspect most people actually want the 'force' behavior as the default and everyone will just do 'marshal build -f' all the time anyway.

Opinions?

a0u:
> This is to avoid wiping out ephemeral data on your rootfs accidentally.

Is this a concern about overwriting a parent's rootfs image that has been modified by the user after build time (e.g., customized through a running system)?  Personally, I just assume that any file generated by marshal can potentially be clobbered anytime I run `marshal build`.  It's usually more confusing to me when something _doesn't_ change as expected through the dependency tree.

Issue #42, opened 2019-07-10
Add deep clean option


NathanTP:
"marshal clean" currently only cleans the top-level workload. It does not affect any dependencies. So if you have a deep chain of workloads that are based off eachother, you have to manually clean all of them (or just go nuclear and "rm images/*").

It would be nice if 'clean' had a '--deep' option that cleaned dependencies all the way down.

PR #182, opened 2020-12-03
Customize Distros

NathanTP:
This is a fairly big overhaul of how distributions are handled. The biggest user-visible change is that they may now provide the 'distro' option which includes a distro-specific 'opts' section for customizing distros. Currently only buildroot has any options to customize. You can see the docs for details but basically users can now provide a custom buildroot configuration (including a custom busybox config) as well as environment variables to pass to the buildroot make command.

The way this works is that if a workload customizes its distro, Marshal will fork the chain of parents to create a distro-specific hierarchy. For example, a normal workload might look like "br -> br-base.json -> example.json", but if it customizes the distro we'd get "br.0DFA -> br-base.json.0DFA -> example.json" (where 0DFA is some marshal-generated identifier for this distro configuration). This will use up some extra disk space, and it will add some build time (since buildroot must be built from scratch taking ~20min). 

A less visible but also important change is that distros are more factored now. There isn't anything in the mainline marshal code that looks for specific distros. This will make it easier to add new distros in the future (it's pretty simple to do now from a marshal perspective).

Backward-Breaking Changes:
- Workloads may no longer base directly on a distro, they must either directly specify the distro using the 'distro' option, or use one of the board's base workloads (recommended). I don't imagine anyone was doing this anyway (you really should use base workloads).
- You can no longer hard-code disk images (at least not without basing on a valid workload). This was a pretty esoteric use-case and was always a hack. It's easy enough to just copy the image you want to images/workload.img.

I'd like for someone to try this out before merging it so we can be sure we hit the needed functionality (@abejgonzalez ?)


wlutil/wlutil.py
```diff
@@ -19,6 +19,7 @@
 import re
 import pprint
 import doit
+import importlib
```
a0u, line 22: Running `marshal build boards/firechip/base-workloads/br-base.json` on Python 3.6.8 throws an error:
```
Traceback (most recent call last):
  File "./marshal", line 218, in <module>
    main()
  File "./marshal", line 73, in main
    wlutil.initialize()
  File "/home/centos/chipyard/software/firemarshal/wlutil/wlutil.py", line 390, in initialize
    ctx = marshalCtx()
  File "/home/centos/chipyard/software/firemarshal/wlutil/wlutil.py", line 242, in __init__
    self.deriveOpts()
  File "/home/centos/chipyard/software/firemarshal/wlutil/wlutil.py", line 332, in deriveOpts
    m = self.importDistro(dPath)
  File "/home/centos/chipyard/software/firemarshal/wlutil/wlutil.py", line 293, in importDistro
    spec = importlib.util.spec_from_file_location(distroPath.stem, distroPath / "__init__.py",
AttributeError: module 'importlib' has no attribute 'util'
```
This happens even with a fresh clone and may be related to this [partial initialization issue](https://stackoverflow.com/questions/39660934/error-when-using-importlib-util-to-check-for-library).

I am able to work around it with
```suggestion
import importlib.util
```

a0u: Oddly enough, the same import was used before without any problems.
https://github.com/firesim/FireMarshal/blob/5848bd3946806f2565165d23058d08b074b430de/wlutil/install.py#L5

NathanTP: ya, Abe just found that one to. It's weird because it works fine on my machine. I think this is an issue with a specific version of importlib (I'm using anaconda, but still python 3.6). Anyway, your suggestion will fix it.

.gitmodules
```diff
@@ -8,7 +8,7 @@
 	path = boards/default/firmware/opensbi
 	url = https://github.com/riscv/opensbi.git
 [submodule "br/buildroot"]
```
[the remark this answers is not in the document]
abejgonzalez: Beat me to it. I have a commit with this fix (and some other README.md) stuff ready to go.

abejgonzalez:
Is it worth mentioning that changing something like a `BUSYBOX` config without changing anything else will not trigger recompilation of things? I know that it is implied but it might be worth it to be explicitly clear.

NathanTP:
Ya, I noticed that and wasn't sure what to do about it. I don't think marshal should start parsing your buildroot configs and it's a more general problem than that. One solution is to add a new 'deps' option to workload configs for you to explicitly add dependent files. I might want to defer that to a later PR though.

boards/default/distros/br/br.py
```diff
@@ -0,0 +1,244 @@
...
+    if len(opts) == 0:
+        return None
+
+    h = hashlib.md5()
+    if 'configs' in opts:
+        for c in opts['configs']:
+            with open(c, 'rb') as cf:
+                h.update(cf.read())
+
+    if 'environemnt' in opts:
```
abejgonzalez, line 61:
```suggestion
    if 'environment' in opts:
```

NathanTP:
As far as I know, there are only two outstanding issues with this PR that Abe ran into during testing:

1. BusyBox configuration changes are not detected by FireMarshal (since they're only referenced by the buildroot config and marshal doesn't do that sort of second-order dependency search).
2. Changes to the base image don't propagate to the workload without explicit cleaning (and even then, without a deep clean option you have to explicitly clean the whole inheritance chain).

Issue 2 is a well-known pre-existing issue (see #42).

Issue 1 is not really a new issue but we work around it by ensuring that host-init scripts are never uptodate (that way they will always pick up changes). I'm sure that's not a foolproof solution though. With the buildroot configs, the issue is more prominent since changing busybox configs is a common use-case. I think the best solution here is to add a 'deps' option to workload configs that lists any additional files that marshal should consider for deciding if it needs to rebuild a workload.

For both issues, I propose solving them in separate PRs since they are more general than just this custom distro thing and aren't show-stoppers.

NathanTP:
Issue 1 is actually already in the issue tracker #43

boards/default/distros/br/br.py
```diff
@@ -0,0 +1,244 @@
...
+            "configs" : base['configs'] + new['configs'],
+            "environment" : {**base['environment'], **new['environment']}
+    }
+
+    return merged
+
+
+def initOpts(cfg):
+    """Given a raw marshal config object, perform any distro-specific
+    intialization (using the cfg['distro']['opts'] field."""
```
abejgonzalez, line 79:
```suggestion
    intialization (using the cfg['distro']['opts'] field)."""
```

boards/default/distros/br/br.py
```diff
@@ -0,0 +1,244 @@
...
+                'img' : self.outputImg
+                }
+
+
+    def configure(self):
+        """Construct the final buildroot configuration for this environment. After
+        calling this, it is safe to call 'make' in the buildroot directory."""
+
+        toolVer = wlutil.getToolVersions()
+        # Contains options specific to the build enviornment (br is touchy about this stuff)
```
abejgonzalez, line 139:
```suggestion
        # Contains options specific to the build environment (br is touchy about this stuff)
```

abejgonzalez (commented):
LGTM minus the small spelling nits.

abejgonzalez (approved):

a0u (approved):
Re-approving.

PR #186, opened 2021-01-23
Yaml

NathanTP:
I'm tired of not being able to comment stuff out. YAML is a bit nicer to work with anyway. I don't anticipate using any fancy features, and JSON will remain supported.


scripts/fullTest.py
```diff
@@ -126,7 +126,7 @@ def runTests(testNames, categoryName, marshalArgs=[], cmdArgs=[]):
 
     for tName in testNames:
         log.log(logging.INFO, "[{}] {}:".format(categoryName, tName))
-        tPath = testDir / (tName + ".json")
+        tPath = testDir / (tName + ".yaml")
```
a0u, line 129: It seems that all `test/*.json` files should be renamed to `*.yaml`, but they don't appear to be committed yet.  I see a bunch of file deletions in the diff but no moves.

NathanTP: derp. fixed

wlutil/config.py
```diff
@@ -1,6 +1,7 @@
 import glob
 import collections
 import json
```
a0u, line 3: Is this import still necessary?

a0u (dismissed):
LGTM.

wlutil/config.py
NathanTP: nope. In fixing that I also found a bug in the exception handler for invalid yaml files. Anyway, both fixed now.

a0u (approved):

PR #189, opened 2021-03-17
Support having base workloads not having modules

abejgonzalez:
I am porting FireMarshal to non-FireSim board targets. In the process, I noticed a bug where the `modules` key wouldn't be found in my config where there are no Linux kernel modules causing an error. This prevent those types of errors.

Sample `br-chip-base.json` tested with:

image

Note that this looks basically the same as the original `br-base.json`. Most of the unique changes are in the `<buildroot/linux>-config`s.


NathanTP (approved):
Seems reasonable

PR #190, opened 2021-04-02
FireMarshal for Prototype Platform

abejgonzalez:
This is the initial work to get FireMarshal working for the prototype platform (specifically the VCU118). This adds a new "board" called `prototype` and a new "simple" configuration called `br-chip-base.json`.

This works with GPIOs, SDCard.

TODO:
- [x] Add `install` functionality to run the `change-addresses` on the output of the `build`
- [x] Clean up the Linux and Busybox configs more
- [x] Fix the `modprobe` failing since there are no Linux modules on the prototype (fails in the `init` script when loading FireMarshal modules and in the default init script)
- [x] Expect FireMarshal test output (from `./scripts/fullTest.py`)


boards/prototype/base-workloads/br-chip-base.json
```diff
@@ -0,0 +1,20 @@
...
+      }
+  },
+  "overlay" : "overlay",
+  "linux" : {
+      "source" : "../../linux",
+      "config" : "linux-config"
+  },
+  "firmware" : {
+      "use-bbl" : false,
+      "bbl-src" : "../../firmware/riscv-pk",
```
NathanTP, line 16: Does BBL work at all or do you just prefer not to use it by default?

boards/prototype/base-workloads/br-chip-base/linux-config
```diff
@@ -0,0 +1,60 @@
+#CONFIG_BLK_DEV_INITRD=y
+#CONFIG_EMBEDDED=y
+#CONFIG_HZ_100=y
```
NathanTP, line 3: Have you confirmed that these are not set in the final config? My understanding is that to disable something that was set by defconfig you have to do: 
`# CONFIG_HZ_100 is not set`

or
`CONFIG_HZ_100=n`

I wonder if I should change marshal to use defnoconfig instead of defconfig and have the base workloads specify way more stuff...

NathanTP (commented):
This looks good to me in general, just some minor comments.

I'll trust you to have tested it sufficiently. Have you run firemarshal/scripts/fullTest.py with the different board configured? I suppose it should work but it's worth checking.

boards/prototype/base-workloads/br-chip-base.json
abejgonzalez: BBL works. When I originally set this up I was toggling between the two to find out why things weren't booting. Are you advocating to clean up things a bit?

boards/prototype/base-workloads/br-chip-base/linux-config
abejgonzalez: I don't think they are getting set (though I haven't explicitly checked). I still need to clean this up though. 

As for moving to `defnoconfig`... that is definitely an option but wouldn't that make it harder to port to newer Linux versions?

boards/prototype/base-workloads/br-chip-base.json
abejgonzalez: FYI, the reason FireMarshal didn't originally work on the FPGA was SDCard space issue (the chip bootrom would load in 20MiB exactly and the FireMarshal payload was slightly larger than 20MiB...)

abejgonzalez:
> This looks good to me in general, just some minor comments.
> 
> I'll trust you to have tested it sufficiently. Have you run firemarshal/scripts/fullTest.py with the different board configured? I suppose it should work but it's worth checking.

I think most of it should work... but I'm not sure if you have tests that need the block/NIC modules. I'll do more thorough testing before it gets merged.

boards/prototype/base-workloads/br-chip-base/linux-config
NathanTP: Ya, the defnoconfig would definitely be annoying which is why I don't do it. So far it looks like we're fine with it as is.

NathanTP:
I think there are one or two tests that rely on the drivers. The test script will run all tests and report which ones failed so we can double check that it's only those that we expect to fail.

NathanTP (approved):
lgtm

abejgonzalez:
As an FYI, Im getting alot of failures running the regressions suite that seem to point to BWRC issues rather than issues with the workload. I've gone ahead and tested the following on the VCU118 prototype platform:

`./marshal -v -d build br-base.json` with BBL and OpenSBI
 and
`./marshal -v -d install -t prototype br-base.json`

I'll run the test suite on an EC2 machine later to see if I can get things to work better.

abejgonzalez:
I don't think running the test suite is good right now mainly because the prototype platform doesn't include `bash` which nullifies most if not all of the tests (things hang since the scripts can't be run). I fixed some other small issues but this should be good to go now if you don't mind running the test suite fully.

NathanTP (approved):
lgtm. All tests passed on firechip board.

PR #191, opened 2021-04-07
Pass environment variables to br kfrag generation

abejgonzalez:
This is used to set an environment variable for `BR2_EXTERNAL` while the `.config` is being created (necessary for external buildroot trees to work). This fixes a bug where if you have a `BR2_EXTERNAL` tree that has its own kfrag element (i.e. `BR2_PACKAGE_MY_SPECIAL_INCLUDE=y`) it doesn't get dropped before doing the main build (because the main build has the variable set while the configuration doesn't).

Reference: https://buildroot.org/downloads/manual/manual.html Section 9.2

### DIFF
--- a/boards/default/distros/br/br.py
+++ b/boards/default/distros/br/br.py
@@ -131,7 +131,7 @@ class Builder:
                 }
 
 
-    def configure(self):
+    def configure(self, env):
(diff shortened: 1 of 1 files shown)


NathanTP (approved):
lgtm

PR #193, opened 2021-06-10
Expand environment variables in buildroot

NathanTP:
This enables users to include variable substitutions when providing additional environment variables to the buildroot build process. These can come from the local environment and the special $WORKLOAD_NAME_PATH variable that marshal sets (so you can reference your own workdir).


a0u (approved):
Thanks for adding this feature.

This is needed for ucb-bar/sha3#30.

PR #197, opened 2021-07-13
Sleep(1) before calling umount

davidbiancolin:
Since updating my manager, i've repeatedly run into an apparent race where marshal tries to unmount the image while there's still lingering activity. I'm not sure if this is the "right" fix, but it's helped me.

### DIFF
--- a/wlutil/wlutil.py
+++ b/wlutil/wlutil.py
@@ -528,6 +528,7 @@ if sp.run(['/usr/bin/sudo', '-ln', 'true'], stdout=sp.DEVNULL).returncode == 0:
         try:
             yield mntPath
         finally:
+            time.sleep(1)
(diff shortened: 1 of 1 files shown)


NathanTP:
This is strange. I have a vague memory of running into this issue before and fixing it but I don't remember the details. I wonder if an explicit filesystem flush would work. I'd rather not add 1 second to every mount (there can be several in a single build). We'd need to test the solution on both centos and ubuntu (sudo mount vs guestmount).

NathanTP:
We could also try adding the 'sync' option to our mounts. Performance of the mounts isn't important so sync is fine:

    mount -o loop,sync

davidbiancolin:
We could also do a lazy unmount?

If you think using sync is the way i can pr that.

PR #200, opened 2021-07-14
Retry unmounting on failure

davidbiancolin:
Subsumes #197.

A second pass.


NathanTP (approved):
Looks good. I can't reproduce the error since it's pretty rare, but this should fix it. If this doesn't fix the error, then something bigger is wrong. Either way, this fix is useful.

PR #201, opened 2021-08-05
Create install_packages.sh

Quizenger:


NathanTP (approved):
This looks fine for now as an intermediate step. I've added you as a contributor so you can just push directly to the qemuNW branch on this repo. We'll do a bigger PR at the end.

PR #203, opened 2021-08-24
Bump fedora

NathanTP:
This bumps fedora to a more minimal image (and more recent build). We had previously been working off an arbitrary full-sized nightly from when they were first porting Fedora. This one seems a bit more stable and definitely much smaller (800MB vs 6GB). Despite the name, it's not all that minimal. For example, it includes Python.

I'm not sure if there are any packages we think should be included by default, but I can add a guest-init for that. I'm also considering adding a guest-init that updates the dnf repository caches to make installing stuff faster for everyone. Since the caches are old it takes a few minutes the first time you install something. Also, this is a good opportunity to update other stuff if we think there's something with Fedora we really need.

BTW: This is still experiencing ucb-bar/chipyard#950 but it boots in normal disk mode.


alonamid (commented):
Is the networking part related to the original intent of this PR?
Otherwise, LGTM. Tagging @jerryz123 and @a0u who occasionally use the fedora image in case they have any other package requirements.

NathanTP:
lol, no. Somehow I got my branches crossed. It should only be the changes to Fedora's makefile. I fixed it.

alonamid (approved):

PR #207, opened 2021-08-24
Set jlevel based on available CPUs

NathanTP:
Marshal used to default to '-j' for parallelism which uses unbounded threads. This never really caused any issues but is also not a great idea. This makes a jlevel of null default to the number of available cores which is much safer.


wlutil/wlutil.py
```diff
@@ -335,6 +335,10 @@ def deriveOpts(self):
             m = self.importDistro(dPath)
             self['distro-mods'][m.__name__] = m
 
+        if self['jlevel'] is None:
+            self['jlevel'] = psutil.cpu_count()
```
timsnyder-siv, line 339: I think there is a case where `psutil.cpu_count()` can return `None` and you should probably handle that by throwing an exception and telling the user to set it in their marshal.yaml.  I still would never use unbounded `-j` because that causes responsiveness problems that you're trying to address here.

timsnyder-siv (changes requested):
I think you probably need to handle the case where `psutil.cpu_count()` returns `None` as documented in https://psutil.readthedocs.io/en/latest/#psutil.cpu_count

wlutil/wlutil.py
NathanTP: fixed, defaults to 1 (and a warning) if cpu_count() fails.

timsnyder-siv (approved):
Looks good now.  Thanks.

PR #211, opened 2021-09-03
Targeted config load

NathanTP:
FireMarshal currently loads every workload config file on its search path. This made it easy to track down 'base's from other workloads but meant we loaded anything with a .json or .yaml extension which caused false warnings and occasionally errors. It was also inefficient, though other optimizations have since made the performance aspect not that big a deal.

This PR changes marshal to only load explicitly requested workloads by following the inheritance chain during config loading.


NathanTP:
Also, this makes base definitions a bit more strict. We used to get away with 'base'ing off of 'bare', but that was technically not allowed (not since we added the 'distro' config option). This PR will make that usage an error which could break some existing workloads. Instead, they should base on 'bare-base.json'. This will be documented in the release notes but the error won't be super clear to users that have upgraded.

wlutil/config.py
```diff
@@ -640,21 +686,42 @@ def __init__(self, dirs=None, paths=None):
...
-                log.warning("\t" + repr(e))
-                del self.cfgs[cfgName]
-                raise
-                continue
+                raise WorkloadConfigError(cfgName,
+                                          extra="Unable to parse config: " + repr(e))
 
             log.debug("Loaded " + str(cfgName))
 
+    def _loadRecursive(self, cfgName):
```
abejgonzalez, line 698: Where is this used? Isn't this just a copy of the loading from above?

NathanTP: good catch. I think it was leftover from some half-finished idea. Removed now.

abejgonzalez (approved):
LGTM

PR #214, opened 2021-10-20
Add hook to run script after base image is created, but before the full image is made

abejgonzalez:
Adding `host-init-post-img` will run a script similar to `host-init` but will run it once the initial image is made (aka the distro folder is configured / built).

Use case:
You are using buildroot and want to cross-compile some program Y. However, Y needs access to libraries that were cross-compiled in the buildroot image. You can use this to run a script after this point (when everything in buildroot is cross-compiled) and get access to its `output` `staging` area.

TODO:

- [ ] Figure out how to pass env. variables pointing to the distro directory. Double-check this works with Fedora as well (aka generalize it).
- [ ] Better naming of the flag
- [ ] Update the documents. Also probably elaborate on when `host-init`, `guest-init`, `post-bin` are run exactly.

### DIFF
--- a/wlutil/build.py
+++ b/wlutil/build.py
@@ -260,6 +260,9 @@ def addDep(loader, config):
         if 'guest-init' in config:
             img_file_deps.append(config['guest-init'].path)
             img_task_deps.append(str(config['bin']))
+        if 'host-init-post-img' in config:
+            img_file_deps.append(config['host-init-post-img'].path)
+            img_task_deps.append(str(config['bin']))
         if 'runSpec' in config and config['runSpec'].path != None:
             img_file_deps.append(config['runSpec'].path)
         if 'cfg-file' in config:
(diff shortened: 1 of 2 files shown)


NathanTP:
I'm not wild about adding another script. I think we might be able to make it so that the guest-init always runs after the parents have been built (will require some fiddling with the doit stuff in build.py). This shouldn't violate anyone's assumptions and the docs only say that it's run before building your workload so it shouldn't break anything.

That being said, the use-case for this is sort of interesting. Marshal makes no promises about where buildroot lives or what it does when building it. While there's no reason to, we could in theory decide to 'make clean' or something in buildroot after copying the binary out. I don't know of a good forward-compatible way of doing what you want.

NathanTP:
I took a look and it should be pretty simple to ensure that host-init runs immediately before the workload (and after all the parents have been built). Just add the bases to the dependencies for host-init in [build.py](https://github.com/firesim/FireMarshal/blob/73f17936e7bd4dcd01175181496bf814b692ac5f/wlutil/build.py#L155):
```
        loader.addTask({
            'name': str(config['host-init']),
            'actions': [(handleHostInit, [config])],
            'task_dep': config['base-deps'],
        })
```

We still have the issue of relying on where exactly buildroot is, but I suppose that's just a risk your workload has to take, it should be an easy fix if marshal ever does change that.

a0u:
> We still have the issue of relying on where exactly buildroot is, but I suppose that's just a risk your workload has to take, it should be an easy fix if marshal ever does change that.

For build systems based on autotools or pkg-config, the cleaner option is to convert the workload into an external buildroot package so the dependency management works without explicitly referencing the buildroot location.  If that's not possible, it should be simple enough for marshal to provide the absolute buildroot path through an environment variable.

NathanTP:
I have confirmed that any host-inits from parents are guaranteed to run before child workloads. This should address the problem. Closing.

PR #222, opened 2021-12-15
Br jlevel

NathanTP:
Use the global jlevel option for the buildroot distro


boards/default/distros/br/.ropeproject/config.py
```diff
@@ -0,0 +1,114 @@
+# The default ``config.py``
```
abejgonzalez, line 1: What is this file supposed to do? Isn't this separate from the other changes?

abejgonzalez (approved):
LGTM

PR #224, opened 2022-01-18
Fixes a bunch of inconsistencies that have crept in in dealing with

NathanTP:
Allows bare-metal workloads to hard-code a disk image. This is useful for debugging the disk driver or doing other low-level things with the disk.


abejgonzalez (approved):
Seems fine to me

PR #228, opened 2022-02-14
Deprecate the use of a 'dev' branch

NathanTP:
The separate 'dev' branch can cause problems/annoyance for little benefit:

1. PRs default to 'master' and people have to explicitly change it to dev each time.
2. Releases are a more complicated process of rc branch->dev, then dev->master.

From now on, we'll PR directly to master which is now the 'unstable' branch. Things should generally work in master (and hopefully we'll get CI setup someday). Stable releases are always tagged and listed in github releases. This matches most github projects that I'm familiar with.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,7 +16,7 @@ name: Lint Code Base
 #############################
 on:
   pull_request:
-    branches: [master, dev]
+    branches: [master]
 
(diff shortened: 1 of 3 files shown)


NathanTP:
ping @abejgonzalez, this is just a docs change

abejgonzalez (approved):

PR #230, opened 2022-02-18
bump bbl

NathanTP:
Bumps BBL to latest master. This is needed for a planned qemu bump inchipyard (old BBL doesn't like new QEMU).


NathanTP:
I've tested this in spike and qemu. I don't have any firesim instances ready, could someone that has a FireSim environment already set up try it? (@sagark?)

a0u (approved):
Tested with firesim/firesim@eb5612627708e5ef15e097513425b4f09791ee2c.

PR #234, opened 2022-03-23
Build failures should eventually cause marshal to exit non-zero

timsnyder-siv:
other callers of marshal will expect the error to be indicated at a high level using
exit code.

I didn't add special messages but I did make it so that launch failing to launch any workload will exit non-zero and in general anything that ends up at the end of the main loop with `failCount !=0` will also exit non-zero.


NathanTP (approved):
Looks good to me

PR #237, opened 2022-03-31
bump buildroot to 2022.02

### LINKED ISSUE #232: Bump buildroot submodule version to pull in upstream fixes
In the process of testing other dependency updates for Chipyard (https://github.com/ucb-bar/chipyard/pull/1120), I discovered some build failures in FireMarshal on Ubuntu 21.10. Currently, the FireMarshal README mentions Ubuntu 18.04, and the FireSim docs mention Centos 7.7, so these dependency version incompatibilities aren't immediately relevant, but this issue captures the relevant details so you have them when you need them.

The first failure has already been reported in #30, and that workaround has to be applied before the subsequent failures will appear.

The second failure is that buildroot fails to build images with errors "You must install 'python' on your build machine". This has already been fixed upstream in buildroot: The buildroot changelog for 2022.02-rc1 notes: "Python 2.7 and python 2.x specific packages removed as python 2.x is EOL since April 2020." (The specific relevant change is https://github.com/buildroot/buildroot/commit/9c0c7846cdb63f3ed62563c4f532ad02ad0a1c98). The short-term workaround for this error is to create a symlink named 'python' pointing to 'python2' (Ubuntu stopped shipping a binary named 'python' years ago, and only ships binaries named 'python3' and 'python2').

The third failure in building images is a C compile error in fakeroot, which is pulled in as a dependency by buildroot. The problem has already been fixed upstream, first by patching buildroot (https://github.com/buildroot/buildroot/commit/f45925a951318e9e53bead80b363e004301adc6f) and later by updating the version of fakeroot that buildroot pulls in as a dependency (https://github.com/buildroot/buildroot/commit/91ebf2b58b119a43134a95994c3cd63297f599fa). 

Bumping the git submodule for buildroot to a new version would resolve the second and third failures, but it needs testing on older versions of Centos to make sure the dependency update doesn't break the current AMIs recommended for use in ec2.

### LINKED ISSUE #236: Bump buildroot to be able to build with make 4.3
buildroot has already fixed upstream so that it builds cleanly with make 4.3.  If we cherry-pick https://github.com/buildroot/buildroot/commit/9e2128bf5072e5f2fd69e2fc0239558782dfc677 it fixes the issue that I hit in firesim CI at https://github.com/firesim/firesim/runs/5773926269?check_suite_focus=true#step:3:58

The quick-fix for this would be to switch back to using firesim/buildroot with only that patch cherry-picked.

timsnyder-siv:
I was able to build br-base.json just fine without any other changes to marshal.  I did run into the 'open files' ulimit.  I don't ever remember running into that and I don't think that firesim's machine-launch-script.sh increases the ulimit at all.  Maybe we should?

I'm running firesim/deploy/regression.sh on the default AGFI's.  @NathanTP if there is anything else I should run to stress test things after bumping buildroot, I'd be happy to.

fixes #232 
fixes #236

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,7 +16,7 @@ name: Lint Code Base
 #############################
 on:
   pull_request:
-    branches: [master, dev]
+    branches: [master]
 
 ###############
 # Set the Job #
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,11 +4,10 @@ Contributing to FireMarshal
 ## Contributing Changes
 We welcome contributions to FireMarshal through [github
 PRs](https://github.com/firesim/FireMarshal/pulls). **All PRs should be against
-the dev branch**. Master is only updated by the FireMarshal maintainers and
-serves as the stable release branch.
+the master branch**.
 
 Before submitting a PR:
-* Merge origin/dev to ensure you have the latest changes.
+* Merge origin/master to ensure you have the latest changes.
 * Run ./fullTest.py to ensure the basic unit-tests still work
 * Ensure all code passes lint (this will be enforced by github). You can
   manually check each file (or a glob) with the following command:
@@ -20,7 +19,6 @@ Before submitting a PR:
 * For bugs and feature requests: use the github issue tracker: https://github.com/firesim/FireMarshal/issues
 
 ## Branch management:
-1) **master**: Stable release. Updated by FireMarshal maintainers as part of a release. 
-2) **dev**: Unstable development. Dev should (hopefully) work at all times, but
-   there are no guarantees. All changes should be PRd against dev. 
-3) **All other branches**: Private development branches, typically not suitable for use.
+1) **tagged releases**: All stable releases are tagged and tracked with githubs releases mechanism.
+1) **master**: Unstable development. Master should work at all times, but there are no guarantees that bugs haven't been introduced.
+2) **All other branches**: Private development branches, typically not suitable for use.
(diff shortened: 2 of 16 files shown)


timsnyder-siv:
I should also note that I was able to build buildroot while using the `use_conda` branch of firesim that is currently in firesim/firesim#986

NathanTP:
The full suite of tests is in scripts/fullTest.py so running that should have pretty good coverage. I'll also make a post on our Chipyard slack channel to make sure everyone's cool with it.

The ulimit thing is annoying. I'm probably not the right person to decide on that either way. My usual approach is to minimize changes to the host environment when possible. Would reducing the JLEVEL in buildroot solve it? I think I currently set it to the number of host CPUs by default.

One hacky thing we've been doing for the tutorial involves linking against some buildroot stuff to build the "John the Ripper" benchmark. @a0u do you have any concerns with this bump for that?

NathanTP:
@a0u @timsnyder-siv 
Sorry I've been out of commission for a while. I'll test this now in my software-only environment (it sounds like it works fine with hardware if it passes FireSim's regressions). What is the status of this? Is anyone opposed?

timsnyder-siv:
FireSim ended up pinning the version of make to not use 4.3 until we had more time to test bumping buildroot https://github.com/firesim/firesim/blob/89ff00224224ef85d9a726843f6ee255d6f82fe6/scripts/machine-launch-script.sh#L255-L256 and it hasn't been a high enough priority for me to get back to it since then.

timsnyder-siv:
@NathanTP I'll create a PR in FireSim that runs the FPGA regression in hardware.  It is minimal but at least boots linux.

NathanTP:
I've tested on my setup and everything looks good. I know @timsnyder is having ulimit issues, I'm not sure what the best way to handle that is or if it should block this PR.

I'll give this my provisional approval.

abejgonzalez (approved):
LGTM. Boots Linux in FireSim as well.

Issue #125, opened 2020-01-27
simArgs test determinism


NathanTP:
The simArgs test started failing because the amount of memory reported by the simulator changed from 3958 to 3956 (the workload requests 4GiB). I'm not sure exactly where that discrepancy comes from (spike and qemu always differed by a few KB).

Ideally, we'd check to make sure memory settings work, but we might have to stick to just checking CPU count.

This is a problem with the test, not the feature. Simargs work correctly.

NathanTP:
Addressed by #241

PR #241, opened 2022-07-13
SimArgs Determinism

NathanTP:
Make the simArgs test only look at CPU. The memory option is unreliable as the exact memory value can vary based on the simulator.

Addresses #125


abejgonzalez (approved):
SFTM

Issue #239, opened 2022-05-24
FireMarshal does not work with the latest doit version


Schmitz48:
Hi, 

I tried to run the `./marshal build br-base.json` command and I got the following error message: 
`NotImplementedError: doit.cmd_base.py:TaskLoader was removed on 0.36.0, use TaskLoader2 instead`

Simply changing the TaskLoader to TaskLoader2 in wlutil/build.py did not work for me so I downgraded my doit version. 

Regards,
Philipp

Schmitz48:
Issue has been addressed :+1:

PR #242, opened 2022-07-14
Newer python support

NathanTP:
Fixes some incompatibilities with more recent versions of Python. 3.6 remains the officially supported version, but 3.10 now works as well.

Compatibility break: the doit package will need to be >=0.34 which is the default for 3.6 in pip now, but some users may still be on an older version and will need to update.

Risks: I updated to a newer API for doit and I can't be 100% sure that all the dependency tracking still works exactly the same way. It's very hard to get coverage on dependency tracking, though I do have a few tests and I manually tried a few things.

Addresses: #239

### DIFF
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -1,5 +1,5 @@
 psutil
-doit
+doit >= 0.34
 gitpython
 humanfriendly
 PyYAML
--- a/wlutil/build.py
+++ b/wlutil/build.py
@@ -11,7 +11,7 @@ from . import launch as wllaunch
 taskLoader = None
 
 
-class doitLoader(doit.cmd_base.TaskLoader):
+class doitLoader(doit.cmd_base.TaskLoader2):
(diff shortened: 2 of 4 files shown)


abejgonzalez (approved):
Generally, LGTM. FireSim is using Python 3.8, will this work with 3.8?

abejgonzalez:
> Yes, it works with 3.8.
> 
> We could consider bumping the official marshal python version to 3.8 in this PR (just a docs change). I don't have any problem with that other than it being unnecessary.

Ok then I'm fine with merging this. I expect Chipyard + FireSim to align on Python versions (soon-ish) and I think the plan was to be on Python 3.8. I'll leave it to your discretion on whether to bump the docs.

NathanTP:
I think I'll leave it as-is for now. No need to rock the boat and it unofficially supports newer versions now which is good enough for me.

PR #249, opened 2022-10-13
WIP: Removing excess rebuild

tianrui-wei:
Currently buildroot gets rebuilt much too often than necessary. In order to clean off old files from buildroot, it just deleted everything including all the host utils, which can get long at times. Instead, we only need to clean the overlay directory we copy things into and the `output/target/*` directory to generate exact up-to-date image.

The caveat here is that we remove stuff in the overlay folder, which may cause problems if we store persistent files there, but that should be discouraged

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/boards/default/distros/br/br.py
+++ b/boards/default/distros/br/br.py
@@ -181,8 +181,8 @@ class Builder:
             # # This is unfortunate but buildroot can't remove things from the
             # # image without rebuilding everything from scratch. It adds 20min
             # # to the unit tests and anyone who builds a custom buildroot.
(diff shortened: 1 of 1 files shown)


tianrui-wei:
Can folks check if `boards/firechip/base-workloads/br-base/overlay/` and `boards/default/distros/br/overlay/` has items that are **NOT** generated at run time? In that case we need to move them to a separate directory and copy into overlay file system at generation.

abejgonzalez (approved):
LGTM. FireMarshal manages the overlay separately from the buildroot image itself so this is fine to do.

PR #252, opened 2022-10-17
Add new chipyard board | fix some other bugs

jerryz123:
- Adds new "chipyard" board, that has a simple no-frills linux/buildroot config optimized for boot speed. This config boots up to 5x faster than the existing firechip config. This config supports only initramfs, and is designed to be a "bare" config to match the default chipyard SoC (with no network or block device).
- Fixes a bug in the br diffconfig script
- Removes initramfs compression by default


boards/chipyard/base-workloads/br-base.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name" : "br-base",
```
a0u, line 2: Identically named workloads result in identically named output images (e.g., `br-base-bin`).
In a future change, it might be a good idea to prefix the filenames with the board name to allow the images to coexist.

wlutil/build.py
```diff
@@ -407,8 +407,6 @@ def generateKConfig(kfrags, linuxSrc):
 def makeInitramfsKfrag(src, dst):
     with open(dst, 'w') as f:
         f.write("CONFIG_BLK_DEV_INITRD=y\n")
-        f.write('CONFIG_INITRAMFS_COMPRESSION=".lzo"\n')
-        f.write('CONFIG_INITRAMFS_COMPRESSION_LZO=y\n')
```
a0u, line 411: For the record, this now defaults to `CONFIG_INITRAMFS_COMPRESSION_GZIP` if left unspecified (which is fine).

wlutil/busybox-config
```diff
@@ -838,27 +838,27 @@ CONFIG_WATCHDOG=y
...
 # CONFIG_FTPD is not set
 # CONFIG_FEATURE_FTPD_WRITE is not set
 # CONFIG_FEATURE_FTPD_ACCEPT_BROKEN_LIST is not set
 # CONFIG_FEATURE_FTPD_AUTHENTICATION is not set
 # CONFIG_FTPGET is not set
 # CONFIG_FTPPUT is not set
 # CONFIG_FEATURE_FTPGETPUT_LONG_OPTIONS is not set
-CONFIG_HOSTNAME=y
-CONFIG_DNSDOMAINNAME=y
+# CONFIG_HOSTNAME is not set
```
a0u, line 860: I suggest reenabling `CONFIG_HOSTNAME` or removing the line in `/etc/inittab` that updates `/etc/hostname`.

It's for an almost purely cosmetic reason, but it does allow the system to boot without visible errors.

a0u (approved):

a0u (approved):

PR #253, opened 2022-10-18
Add example linux-poweroff test

jerryz123:


example-workloads/linux-poweroff.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name" : "linux-poweroff",
+    "base" : "br-base.json",
+    "overlay" : "overlay",
+    "command" : "/poweroff.sh"
```
abejgonzalez, line 5: 2 things:
1. `command` automatically runs `poweroff` after the "command" is run so the `poweroff` in the script in unnecessary. 
2. Why not just run the `echo` directly from here so that there is no need for an overlay + script?

abejgonzalez (changes requested):

abejgonzalez (approved):

PR #254, opened 2022-10-21
Add CI

abejgonzalez:
Testing adding CI

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,0 +1,97 @@
+name: run-tests
+
+on:
+  # run ci when pring to following branches (note: ci runs on the merge commit of the pr!)
+  pull_request:
+    branches:
+      - master
+
+env:
+  REMOTE_WORK_DIR: /scratch/buildbot/firemarshal-ci-shared/firemarshal-${{ github.sha }}
+
+jobs:
+  cancel-prior-workflows:
+    name: cancel-prior-workflows
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Cancel previous workflow runs
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
+
+  # Set up a set of boolean conditions to control which branches of the CI
+  # workflow will execute This is based off the conditional job execution
+  # example here: https://github.com/dorny/paths-filter#examples
+  change-filters:
+    name: filter-jobs-on-changes
+    runs-on: ubuntu-20.04
+    # Queried by downstream jobs to determine if they should run.
+    outputs:
+      run-core: ${{ steps.filter.outputs.all_count != steps.filter.outputs.non-core-files_count }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            all:
+              - '**'
+
+            # Check for if docs files changed
+            docs: &docs-filter
+              - 'docs/**'
+              - '.readthedocs.yml'
+
+            # Check for non-FireMarshal core file changes
+            non-core-files:
+              - *docs-filter
+              - '**/*.md'
+              - '**/.gitignore'
+
+  setup-repo:
+    name: setup-repo
+    needs: change-filters
+    if: needs.change-filters.outputs.run-core == 'true'
+    runs-on: self-hosted
+    steps:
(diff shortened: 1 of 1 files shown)


abejgonzalez:
Currently, there are issues with running a `screen` session. See "Must be connected to a terminal" in https://github.com/firesim/FireMarshal/actions/runs/3298904128/jobs/5441742036#step:3:70

scripts/fullTest.py
```diff
@@ -130,8 +133,8 @@ def runTests(testNames, categoryName, marshalArgs=[], cmdArgs=[]):
 
         try:
             # These log at level DEBUG (go to log file but not stdout)
-            wlutil.run([marshalBin] + marshalArgs + ['clean', tPath], check=True)
-            wlutil.run([marshalBin] + marshalArgs + ['test'] + cmdArgs + [tPath], check=True)
+            wlutil.run([marshalBin] + marshalArgs + ['clean', tPath], check=True, shell=True)
```
abejgonzalez, line 136: This seems to cause `argparse` to fail in the downstream marshal call.

abejgonzalez:
For more context:

This PR is adding CI to run FireMarshal's `fullTest.py` script which should run all the default FireMarshal test suite. Under the hood this `fullTest` script is running `marshal test <...>` for a variety of workloads found in `test/`. The current issue is that this `marshal test <...>` command creates a screen session (see ref 1) that can't run in the current GH-A shell. This is the error that is seen: https://github.com/firesim/FireMarshal/pull/254#issuecomment-1287239154. I don't know if this issue is due to needing to call `subprocess.POpen` with the `shell=True` flag or I need to somehow change the default shell of GH-A to allow for screen sessions.

Ref 1: https://github.com/firesim/FireMarshal/blob/d38234c2e17aaa87d10a10b312b380a1a8e89257/wlutil/launch.py#L142-L145

tianrui-wei:
@abejgonzalez The PR should be working now, thanks for the patches

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+    name: run-tests
+    needs: [setup-repo]
+    runs-on: self-hosted
+    steps:
+      - name: Run tests
+        run: |
+          cd ${{ env.REMOTE_WORK_DIR }}
+          eval "$(conda shell.bash hook)"
+          conda activate $PWD/.conda-env
+          ./scripts/fullTest.py -c baremetal
```
abejgonzalez, line 105:
```suggestion
          ./scripts/fullTest.py
```

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,120 @@
...
+
+  cleanup:
+    name: cleanup
+    needs: [setup-repo, run-tests]
+    runs-on: self-hosted
+    if: ${{ always() && contains(join(needs.*.result, ','), 'success') }}
+    steps:
+      - name: Delete repo copy and conda env
+        run: |
+            echo "Don't delete"
```
abejgonzalez, line 120:
```suggestion
           rm -rf ${{ env.REMOTE_WORK_DIR }}
```

abejgonzalez:
> @abejgonzalez The PR should be working now, thanks for the patches

If you have the time, parallelizing the `fullTests.py` script would be nice

tianrui-wei:
From what I can see at the moment, the parallisation of fulltest.py fails bc of the way firemarshal handles workdir. Because there's only one workdir, it's very easy to have conflicts between different jobs if they're run in parallel

tianrui-wei:
You can see the changes at https://github.com/firesim/FireMarshal/tree/use-conda-fix

abejgonzalez:
Are we sure that this doesn't rebuild br-base for every single job (that shares the same `br-base` spec)(ditto for fedora)? I think that the current repo setup overbuilds things. Instead we should do the following:

1. Read the parent specification of the current workload yaml
2. If the parent specification image was built, then just copy the image into a new name (adding a hash for the specification contents) and copy the workload files into that image.
3. If the parent specification image is not built, build it, and name the image with a hash of the specification (and maybe files used i.e. hash of git repos). Then do step 2.

Depending on the time I have in the next week I can take a stab at this / look into if things are properly cached.

abejgonzalez:
In the interim, I'll just run the bare-metal tests (as a sanity check) while we figure out the situation w/ parallelization/extra-builds.

jerryz123 (approved):

PR #256, opened 2022-11-29
Misc. Feature Dump

abejgonzalez:
This PR does a couple of things:
- Creates a folder for each img/bin output (and adds associated metadata files like the Linux config to it)
- Moves some "Applying..." output to the debug logger
- Images are deleted then re-generated when the base image (the image that is copied and modified) is modified
- Add a new `out-dir` config item that specifies the name of the output directory to put the img/bin into (basically `<image-dir>/<out-dir>/{*.img,*.bin}`)
- Fixes an exception not found error when doing FireSim installs and the path is incorrect
- Adds a new action called `print_deps` that prints out the items that forced a rebuild (useful for debugging).

This was tested with (in all cases things worked nicely - things weren't rebuilt unnecessarily):
- Protobuf benchmark suite
- My personal FireMarshal workloads
- fullTest.py (except for the tests that stall in Spike - due to another issue w/ Linux)

### DIFF
--- a/boards/default/distros/br/br.py
+++ b/boards/default/distros/br/br.py
@@ -121,7 +121,8 @@ class Builder:
         else:
             self.name = 'br'
 
-        self.outputImg = wlutil.getOpt('image-dir') / (self.name + ".img")
+        self.outputDir = wlutil.getOpt('image-dir') / self.name
+        self.outputImg = self.outputDir / (self.name + ".img")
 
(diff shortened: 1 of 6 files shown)


sagark (approved):
LGTM

tianrui-wei (approved):
LGTM, sorry for the delay

PR #259, opened 2022-12-05
Parallelize Firemarshal, Cache CI Runs, Use Conda Lock, Local CI testing

tianrui-wei:
Currently it does the following things:
1. support running actions with act
2. use cached conda for dependency
3. cache git clone and builds
4. try to parallize builds

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

### DIFF
--- a/.github/workflows/run-tests.local.yml
+++ b/.github/workflows/run-tests.local.yml
@@ -1,0 +1,123 @@
+name: run-tests
+
+on:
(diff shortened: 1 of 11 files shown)


abejgonzalez:
Closing since #277 should speed up FireMarshal for most users. We can re-open this later if necessary.

PR #260, opened 2023-01-19
simulation_outputs support for marshal install

joey0320:
Output `simulation_outputs` for firesim workload json file.

### DIFF
--- a/boards/default/installers/firesim/firesim.py
+++ b/boards/default/installers/firesim/firesim.py
@@ -65,8 +65,12 @@ def install(targetCfg, opts):
             if 'outputs' in jCfg:
                 wls[slot]["outputs"] = [f.as_posix() for f in jCfg['outputs']]
 
-            if 'simulation_outputs' in jCfg:
-                wls[slot]["simulation_outputs"] = [f.as_posix() for f in jCfg['simulation_outputs']]
+            if 'firesim' in jCfg:
+              if 'simulation_outputs' in jCfg['firesim']:
+                wls[slot]["simulation_outputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_outputs']]
+              if 'simulation_inputs' in jCfg['firesim']:
+                wls[slot]["simulation_inputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_inputs']]
+
         fsCfg['workloads'] = wls
     else:
         # Single-node run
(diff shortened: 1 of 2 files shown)


jerryz123:
What's an example of how you use this?

abejgonzalez:
For future reference @joey0320 you should try to add more explanation in the PR description so that others have some context:

Right now this PR is setting up FireMarshal to be able to target FireSim workloads better by passing through FireSim specific arguments:

Take the following example: https://docs.fires.im/en/latest/Advanced-Usage/Debugging-and-Profiling-on-FPGA/TracerV-with-FlameGraph.html#modifying-your-workload-description

```json
{
  "benchmark_name"            : "linux-uniform",
  "common_bootbinary"         : "br-base-bin",
  "common_rootfs"             : "br-base.img",
  "common_simulation_inputs"  : ["br-base-bin-dwarf"],
  "common_outputs"            : ["/etc/os-release"],
  "common_simulation_outputs" : ["uartlog", "memory_stats*.csv", "TRACEFILE*"],
  "post_run_hook"             : "gen-all-flamegraphs-fireperf.sh"
}
```

Right now FireMarshal can fill everything but the `*simulation_<inputs/outputs>` fields. This is annoying because the normal flow for profiling is:

1. Build your workload in FireMarshal
2. Install (aka create/override a FireSim `.json`)
3. Run test

However, if you want to profile you have to manually modify the `.json` after the installation. But what happens if you want to add more simulation outputs (i.e. for TracerV) and then want to rebuild your SW so you can trigger TracerV on the specified region of the program? In this case, you would override your `.json` on the 2nd install losing your modifications and potentially wasting 1 simulation run.

This PR aims to avoid issues like this by having FireMarshal generate the collateral needed for the backend simulator (i.e. FireSim). The new schema would look like this:

```yaml
{
  "name" : "command",
  "base" : "br-base.json",
  "command" : "echo Global: command",
  "firesim" : {
    "simulation_outputs": [ MY LIST ],
    "simulation_inputs": [ MY LIST ]
  }
}
```

In the case that someone is installing to not-FireSim, the entire "firesim" key would be dropped and unused.

jerryz123 (approved):
Ok. That makes sense to me. I was wondering if the outputs/inputs fields could be derived from the other fields, but it looks like they indeed are very firesim-specific.
Thanks @joey0320

abejgonzalez:
Have you validated that this works in FireSim? Do you need to copy the simulation inputs from the FireMarshal workload into the FireSim workload (maybe you need to add a sim. link). I'm not sure that the FireSim-side of this works yet.

abejgonzalez:
You also need to add documentation on the features so that this is discoverable to other users.

PR #261, opened 2023-01-24
Copy simulation_inputs | add docs

joey0320:
Address comments on #260.
- Copy `simulation_inputs` into the workload slot directory.
- Add documentation.

### DIFF
--- a/boards/default/installers/firesim/firesim.py
+++ b/boards/default/installers/firesim/firesim.py
@@ -69,7 +69,9 @@ def install(targetCfg, opts):
                 if 'simulation_outputs' in jCfg['firesim']:
                     wls[slot]["simulation_outputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_outputs']]
                 if 'simulation_inputs' in jCfg['firesim']:
-                    wls[slot]["simulation_inputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_inputs']]
+                    wls[slot]["simulation_inputs"] = [os.path.basename(f.as_posix()) for f in jCfg['firesim']['simulation_inputs']]
+                    for f in jCfg['firesim']['simulation_inputs']:
+                      os.system(f"cp {f} {os.path.join(fsTargetDir, os.path.basename(f))}")
 
         fsCfg['workloads'] = wls
     else:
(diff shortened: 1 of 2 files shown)


docs/source/workloadConfig.rst
```diff
@@ -296,6 +296,25 @@ should be absolute with respect to the workload rootfs. Files will be placed
 together in the output directory. You cannot specify the directory structure of
 the output.
 
+simulation_outputs
+^^^^^^^^^^^^^^^^^^^^^
+A list of simulation output files to copy after running firesim. Each path
+should be absolute with respect to the `sim_slot_*` directory.
```
abejgonzalez, line 302: - Nit: FireSim instead of firesim.
- I think this documentation exposed something I didn't see before. Right now sim. inputs are from the marshal root directory. However, I think there might be cases in the future where things are taken from the `workdir` (relative to the `workdir`). I think a good extension of this would be to enable inputs from multiple sources: 1. relative to the marshal root directory 2. relative to the `workdir` 3. absolute path to anywhere. To DRY this out there can be a shared func that takes list of locations to look for, list of files to find (both absolute and relative) and returns the absolute paths of all files. Then you can take the output of this function to pass into some copy loop (and/or error).

abejgonzalez: - Nit: We should probably have a new section/header of the docs with simulator-specific flags.

boards/default/installers/firesim/firesim.py
```diff
@@ -69,7 +69,9 @@ def install(targetCfg, opts):
                 if 'simulation_outputs' in jCfg['firesim']:
                     wls[slot]["simulation_outputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_outputs']]
                 if 'simulation_inputs' in jCfg['firesim']:
-                    wls[slot]["simulation_inputs"] = [f.as_posix() for f in jCfg['firesim']['simulation_inputs']]
+                    wls[slot]["simulation_inputs"] = [os.path.basename(f.as_posix()) for f in jCfg['firesim']['simulation_inputs']]
+                    for f in jCfg['firesim']['simulation_inputs']:
+                      os.system(f"cp {f} {os.path.join(fsTargetDir, os.path.basename(f))}")
```
abejgonzalez, line 74: Lets use `shutil.copy` (or one of its variants).

abejgonzalez (changes requested):

docs/source/workloadConfig.rst
joey0320: Do you think the `workdir` thing will be okay in cases when the user specifies a filename & there are multiple files within the search scope with the same name?

abejgonzalez: I think if there are multiple files then there should be an error instead of silently resolving to a specific file.

joey0320: I'll get back to this a bit later :)

joey0320: My impression of the path thing is that the user should know & specify the path marshal should search for each input file. So instead of searching for random locations, shouldn't we group each input file by the base path we want to search for?
Something like this?
```
"simulation_inputs" : {
  "marshal" : [ "input1", "input2"],
  "workload" : [ "input3", "input4"]
}
```

abejgonzalez: I think this adds more verboseness to the file than needed. 

Maybe we can simplify this to be... search in:
- image dir that is generated for the workload
- work dir that is specified
- absolute paths

Why do you need access to the top-level marshal dir?

joey0320: I don't need access to the top-level marshal dir. Rather, I thought it would be cleaner to explicitly specify where to copy files from & for consistency in the above example. I guess, I kind of don't like the idea of searching for files for a set of relative paths. I think absolute paths would be great though.

PR #268, opened 2023-03-22
Repo clean | Add more br packages

abejgonzalez:
Fixes `.gitignore` files to properly ignore generated files. Also, adds `memcached zeromq zmqpp cppzmq` needed for FireSim paper results to be generated (https://github.com/firesim/mutilate-loadgen-riscv-release). I don't know what `BR2_COMPILER_PARANOID_UNSAFE_PATH=y` is, maybe we should unset?


abejgonzalez:
@jz this at least builds in FireSim. this should be good for review.

abejgonzalez:
According to this (below) the extra BR config throws an error if things are misconfigured (which I think is probably good): https://github.com/rancher/buildroot/blob/881dae3a9c086ee3880834e10f72baea221f4945/Config.in#L662

tianrui-wei (commented):
I noticed you're generating these configuration changes with a different buildroot version. Can you include the buildroot bump in your PR as well?

jerryz123 (approved):

abejgonzalez:
> I noticed you're generating these configuration changes with a different buildroot version. Can you include the buildroot bump in your PR as well?

I used the `buildroot` version that was in FireMarshal main. I'll double check again.

jerryz123:
Actually good catch @tianrui-wei ... abe can you regen those with the current BR

jerryz123:
Ah you are right Abe, I didn't do the right thing when I bumped buildroot.

PR #270, opened 2023-03-25
add buildroot example

joey0320:
@charleshong3 
From the marshal directory (that is installed as a CY submodule), if you run `./marshal build example-workloads/example-br.json` it will build a image for you.
After the image is built, run `./marshal install config example-workloads/example-br.json`. This will add a json file to `$CY/sims/firesim/deploy/workloads`.
I guess this will give you an idea on what you have to do lol.

### DIFF
--- a/example-workloads/example-br.json
+++ b/example-workloads/example-br.json
@@ -1,0 +1,13 @@
+{
+  "name" : "example-br",
+  "base" : "br-base.json",
+  "overlay" : "overlay",
+  "host-init" : "host-init.sh",
+  "jobs" : [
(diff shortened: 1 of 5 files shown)


abejgonzalez:
If you want to create a "tutorial" page in the FireMarshal docs then we can add this.

joey0320:
Sure, should I add this before the release?

abejgonzalez:
We can do this after the release. If this is a tutorial, it should be more in-depth and probably have an example of how to compile things for baremetal + Linux in the same setup.

PR #277, opened 2023-06-07
Cached Buildroot Images

abejgonzalez:
This adds support in FireMarshal buildroot to use a cached buildroot image downloaded from the internet. This only happens if all of the following is true:

- The file is successfully fetched from the internet
- No file modifications were made (`changed` var in the code)
- The buildroot image doesn't exist (`task.targets[0]` doesn't exist)
- The buildroot repo isn't dirty (`dirty` dict value in the code)

This also adds 2 CI hooks to build and upload the images:
- A weekly build +upload of the image (so things get updated just in case)
- Whenever `master` is pushed (i.e. after a merge), the CI workflow will re-run and there will be a build + upload of the image

I use https://github.com/firesim/firemarshal-public-br-images as the storage repo.

### DIFF
--- a/boards/default/distros/br/br.py
+++ b/boards/default/distros/br/br.py
@@ -62,7 +62,7 @@ exit""")
 
 
 def hashOpts(opts):
-    """Return a unique description of this builder, based on the provided opts"""
+    """Return a unique description of this builder, based on the provided opts and sha of buildroot"""
 
     if len(opts) == 0:
         return None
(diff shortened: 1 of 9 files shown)


sagark (approved):

boards/default/distros/br/br.py
```diff
@@ -164,11 +189,43 @@ def configure(self, env):
         wlutil.run([mergeScript] + kFrags, cwd=(br_dir / 'buildroot'), env=env)
 
     # Build a base image in the requested format and return an absolute path to that image
-    def buildBaseImage(self):
+    def buildBaseImage(self, task, changed):
```
jerryz123, line 192: Can you add a type annotation and a docstring on what task, changed are? I also couldn't find where this is being passed into this function

abejgonzalez: Sure. They are passed by `doit` to the function. I can add some doc on this.

sagark (approved):

jerryz123 (approved):
Thanks. LGTM now

PR #278, opened 2023-06-13
Use bar-tender specific CI access

abejgonzalez:
Switches CI to use the SSH key and API token given by the `bar-tender` bot.


abejgonzalez:
Subsumed by #280

PR #281, opened 2023-07-12
fix for read cycle issue

SeahK:


jerryz123:
For more context, enabling RISC-V PMU features in linux means that the counters start disabled (via the `mcountinhibit` CSR), and `perf_event_start` syscalls are required (either in user code, or through a utility like `perf`) to access them.

For convenience, disable the PMU features so we can just `rdcycle` in user-code.

jerryz123 (approved):

jerryz123 (approved):

PR #282, opened 2023-07-13
Update fedora-base linux-config

ktpedre:
Enable the SiFive serial port so console output works on FireSim. Before this change, FireSim simulations using the Fedora image had a non-functioning Linux console, with no output after the OpenSBI messages.


abejgonzalez:
Ping @jerryz123

jerryz123:
Oops. Sorry. Must have missed this when testing because spike uses a ns16550 device, rather than the sifive serial implementation. There is some ongoing work towards enabling spike to model the same uart device as the hardware. That would prevent this kind of mismatch in the future.

jerryz123 (approved):

PR #283, opened 2023-07-14
Tutorial docs

joey0320:


docs/source/Tutorials/custom.rst
```diff
@@ -110,6 +110,76 @@ known good output of "Hello World!".
 
 .. todo:: Add (or link to) more detailed examples of bare-metal and rocc workloads.
 
+
+Using the same directory structure for both bare-metal and buildroot
+-----------------------------------------------------------------------
```
abejgonzalez, line 115: Nit: Newline after this.

docs/source/Tutorials/custom.rst
```diff
@@ -110,6 +110,76 @@ known good output of "Hello World!".
...
+Finally the workload consists of individual ``jobs``. The ``outputs`` field
+specifies the files that you want to copy out from the file system after the
+simulation has ended. ``run`` sets the script to run on the target once the
+simulation starts. If you take a look into ``example-br/run.sh``, you can see
+that it is executing our "Hello World" binary and terminating the simulation by
+runnint ``poweroff``.
+
+.. include:: example-workloads/example-br/run.sh
+   :code: bash
+
```
abejgonzalez, line 152: Nit: Consistent newlines (use 2 or 1 between paragraphs)

docs/source/Tutorials/custom.rst
```diff
@@ -110,6 +110,76 @@ known good output of "Hello World!".
 
 .. todo:: Add (or link to) more detailed examples of bare-metal and rocc workloads.
 
+
+Using the same directory structure for both bare-metal and buildroot
```
abejgonzalez, line 114: Looking at the other tutorials here (https://firemarshal.readthedocs.io/en/latest/Tutorials/custom.html), it looks like we really just need to add one on BuildRoot and not baremetal. Even then, the Fedora tutorial describes much of what you want... so I think if the docs really want to get fixed I would do the following:

1. Re-order the "Custom Workloads" tutorials to be 1. BuildRoot Linux Hello World (w/ descriptions of parameters probably taken from the existing Fedora tutorial), 2. Baremetal Hello World 3. Fedora Hello World. 

2. For sharing, I think the `test/bare.yaml` workload already implements the baremetal hello world... instead of copying that we should repurpose it for the tutorial (I think that workload compile could be simplified). OR we can symlink the example from the test area (or vice version).

docs/source/Tutorials/custom.rst
```diff
@@ -110,6 +110,76 @@ known good output of "Hello World!".
...
+.. include:: example-workloads/example-br/run.sh
+   :code: bash
+
+
+We are ready to build and install the workload for FireSim!
+
+::
+
+  ./marshal -v build example-workloads/example-br.json
+  ./marshal -v install example-workloads/example-br.json
```
abejgonzalez, line 159: I would add this as a `.. note::` since FireMarshal is separate from FireSim in the documentation.

example-workloads/example-br/overlay/root/hello-world/Makefile
abejgonzalez, line 1: This Makefile only works for baremetal not Linux since you are using the `unknown-elf` compiler.

abejgonzalez (changes requested):

PR #295, opened 2024-02-06
buildroot perf support

joonho3020:
Add a example workload that copies the `perf` binary and the corresponding dynamically linked libraries into the br image.
This is a workaround as adding a buildroot perf package requires us to use the buildroot kernel.


jerryz123:
Committing compiled objects into this repo makes me nervous. What happens if we bump the kernel? How would we update this?

joonho3020:
That is a good point. But we don't really update that often :)

jerryz123:
> That is a good point. But we don't really update that often :)

But we shouldn't make it harder for ourselves to upgrade....

abejgonzalez (commented):
Whats the current way to add `perf` to `buildroot`? Isn't it just setting `BR2_PACKAGE_PERF`? We can just set that as the default and upload a br-image for that? I agree with Jerry, uploading artifacts of a build IMO isn't clean... even assuming it was there is no script to generate the files when bumping again.

PR #304, opened 2024-06-15
Use FireSim sudo scripts if available

abejgonzalez:
Try to use FireSim's `sudo` scripts when available (instead of defaulting to `guestmount` or using `cpio` with `sudo`).

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -36,7 +36,7 @@ jobs:
       # Checkout the code base #
       ##########################
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           # Full git history is needed to get a proper list of changed files within `super-linter`
           fetch-depth: 0
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -86,9 +86,9 @@ jobs:
           cd ${{ env.REMOTE_WORK_DIR }}
           eval "$(conda shell.bash hook)"
           conda activate $PWD/.conda-env
-          conda env list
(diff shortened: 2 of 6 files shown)


jerryz123 (approved):
This global stuff is ugly... but I don't think I can think of another solution quickly.

abejgonzalez:
I'm unsure why the `fed-smoke0` test doesn't work (it runs locally) and all other tests pass. This should be good to merge.

jerryz123:
Does the CI have the same sudo settings as your machine? I see:
```
2024-06-17T22:07:36.2258753Z 2024-06-17 15:05:55,794 [run         ] [DEBUG]  DEBUG: cp: cannot create regular file '/scratch/buildbot/firemarshal-ci-shared/firemarshal-a21bc92476b0e8375efbee3bdc86ee567e97de48/disk-mount/etc/shadow': Permission denied
``` 
This feels like a related failure

jerryz123 (dismissed):
I didn't realize the tests had been broken... if you judge the CI failures to be unrelated go ahead and merge

abejgonzalez:
Copying to/from the mountpoint is now more complicated. Now it recursively `chmod`'s all directories in the mountpoint s.t. it can safely copy in/out files (these `chmod`'s are reverted afterwards to keep the same permissions on the directories). This works around an issue where `marshal build`'s `mount` would leave a sub-dir in the mountpoint with incorrect permissions while running a separate `subprocess.run(... mount ...)` (virtually the same way FireMarshal does the mount) has the proper permissions on the sub-dir.

jerryz123 (approved):
Thanks for dealing with this

Issue #318, opened 2025-06-27
Fedora Build error


ghost:
The error occurs during the FireMarshal build of the fedora-base workload. After successfully loading the configuration and preparing the Fedora root filesystem image, the process attempts to apply the overlay. Early in the log, there is a warning:

    DEBUG: Failed checking if running in CYGWIN due to: FileNotFoundError(2, 'No such file or directory').

This indicates that FireMarshal tried to detect a CYGWIN environment but could not find the relevant file or directory. While this warning does not halt the build, it may signal missing dependencies or an incomplete environment setup.

The main failure happens when FireMarshal tries to mount the generated fedora-base.img using a loop device to apply the overlay. The mount command fails with the message:

    mount: .../disk-mount: wrong fs type, bad option, bad superblock on /dev/loopXX, missing codepage or helper program, or other error.

The logs confirm that the image file is being treated as corrupted or invalid at the mounting stage, likely due to a problem with the image creation or an incompatibility with the expected filesystem type. The error is reproducible and occurs after several overlay and dependency steps, specifically at the point of mounting the .img file for overlay application.

"../../../firesim/sw/firesim-software/images/firechip/fedora-base" The .img file in this path seems to be corrupted while building.

PR #320, opened 2025-07-31
Change Fedora Rawhide URL to its new location under archive/

ravenjoad:
Sometime in the past 8-12 months, Fedora moved their virt-builder image for Fedora Rawhide 20191123.n.1 under an archive/ URL. Update the Makefile to reflect this change.


ravenjoad:
It would be good if we actually verify against the checksum too. This would prevent us from getting responses or file that we don't expect. (https://dl.fedoraproject.org/pub/alt/risc-v/archive/repo/virt-builder-images/images/Fedora-Minimal-Rawhide-20191123.n.1-sda.raw.xz.CHECKSUM)

With the old URL, curl goes for ~1 minute, then an HTTP response is given and /that/ is saved to the compressed img file. That compressed image is then "decompressed" to a zero-byte file, allowing the rest of Firemarshal to go.

ravenjoad:
Fixes & closes #318.

